PR #272, opened 2015-01-06
Feature/obiee csv export

kerin:
@r4vi @marcofucci opinions please?


kerin:
retest this please

kerin:
retest this please

cla_backend/apps/core/fixtures/test_auth_clients.json
```diff
@@ -84,6 +97,24 @@
         }
     },
     {
+        "pk": 95,
```
r4vi: will this do something weird on prod?

kerin: this fixture doesn't get applied on prod - the API client and user will get manually created in prod instead.

cla_backend/apps/call_centre/views.py
```diff
@@ -517,3 +527,71 @@ def get_queryset(self, *args, **kwargs):
...
+
+    filename = 'cla_database.zip'
+
+    def get(self, request, format=None):
+        """
+        dt_from, dt_to: ISO 8601 datetime string (2014-08-29T23:59:59)
+        passphrase: diversity GPG private key passphrase
+        """
+        dt_from = request.QUERY_PARAMS.get('from', '')
+        dt_to = request.QUERY_PARAMS.get('to', '')
```
marcofucci: would it be better to validate the params and return error codes if invalid? Worried that we would get many sentry errors for params not properly formatted/passed.

kerin: yep, fair point - I was being lazy there

cla_backend/apps/call_centre/views.py
```diff
@@ -517,3 +527,71 @@ def get_queryset(self, *args, **kwargs):
...
+        passphrase = request.QUERY_PARAMS.get('passphrase', '')
+
+        export_path = tempfile.mkdtemp()
```
r4vi: what if it was run twice? maybe worth dumping a ```cla_database.zip.lock``` file;
if call is made again then don't run unless that file is not there, delete the file at the end of the method or when an error happens. Maybe a failsafe to allow running if the ```cla_database.zip.lock``` is older than 1hour?

kerin: Is that necessary? [mkdtemp](https://docs.python.org/2/library/tempfile.html#tempfile.mkdtemp) guarantees no race conditions, so running this twice in quick succession would just mean you get the same file twice. Rate limiting might be an idea, although the actual DB query/output is pretty small I think.

r4vi: yeah my worry was some sort of accidental ddos, maybe rate limit is better here

cla_backend/apps/core/fixtures/test_auth_clients.json
r4vi: :racehorse: :+1:

cla_backend/apps/call_centre/views.py
kerin: Yep, rate limit is a good idea - this method shouldn't be hit more than once a day anyway.

cla_backend/apps/call_centre/sql/ExportStandardTables.sql
```diff
@@ -0,0 +1,30 @@
+CREATE OR REPLACE FUNCTION cla_db_to_csv(dt_from TEXT, dt_to TEXT, path TEXT) RETURNS void AS $$
```
r4vi: take it this is gonna be installed manually too, maybe add it as a migration?

requirements/base.txt
```diff
@@ -26,6 +26,8 @@ jsonpatch==1.7
 networkx==1.9
 lxml==3.3.5
 drf-extensions==0.2.5
+httpsig==1.1.0
+djangorestframework-httpsignature==0.2.1
```
marcofucci: is this the only available DRF package? Seems quite new so maybe not mature enough.

cla_backend/apps/call_centre/sql/ExportStandardTables.sql
kerin: executing this every time was just easier, but yes, could be installed via a migration 😒

requirements/base.txt
kerin: this seems to be the one blessed by DRF itself, and I haven't seen any others. It's about a year old, which seems reasonable to me?

cla_backend/apps/call_centre/sql/ExportStandardTables.sql
r4vi: oh if it runs every time then its fine; prefer that to a migration

keep it like it is

kerin:
@marcofucci @r4vi HTTP error codes added, rate limit added, individual SQL now used instead of a function to dump out everything, and basic sanity test to check that the required tables/columns haven't been dropped... so, opinions please? 😒

r4vi:
image

cla_backend/apps/call_centre/sql/ExportLegalAidEligibilityCheck.sql
```diff
@@ -0,0 +1,23 @@
...
+id,
+created,
+modified,
+reference,
+category_id,
+you_id,
+partner_id,
+disputed_savings_id,
+your_problem_notes,
+notes,
```
marcofucci: Isn't this potentially sensitive? If Yes, maybe we should delete this and eventlog.log.patch as well. What do you think @r4vi ?

cla_backend/apps/call_centre/sql/ExportProviderStaff.sql
```diff
@@ -0,0 +1,13 @@
+COPY (SELECT
+id,
+created,
+modified,
+user_id,
+provider_id,
+is_manager,
+chs_organisation,
+chs_user,
```
marcofucci: not actually sure they need `chs_organisation` and `chs_user`, maybe we can delete them

cla_backend/apps/call_centre/sql/ExportLegalAidEligibilityCheck.sql
kerin: notes and eventlog.patch are sensitive, so I'll remove those - the rest should be ok though

PR #305, opened 2015-02-20
Remove help text in Scope diagnosis graphml

tyom:
[PT](https://www.pivotaltracker.com/story/show/88786744)

Path:
* Education
* Any other education problem
* The elient's problem is about admissions
* Any other issue


andyhd:
@Racketyrac has merged this change into develop already

PR #309, opened 2015-03-23
Prevent "User is not authenticated" error on CLA Public contact page

andyhd:


cla_backend/apps/legalaid/views.py
```diff
@@ -177,7 +178,10 @@ def create_means_test_log(self, obj, created):
         except Case.DoesNotExist:
             return
 
-        user = self.request.user
+        if self.request.user:
```
marcofucci: I would personally implement a `get_user` method on the `legalaid.views.BaseEligibilityCheckViewSet` which defaults to `return self.request.user` and then override that method in `checker.views. EligibilityCheckViewSet` to return `get_web_user()`.

This way, we could catch potential errors when the user is not authenticated but it should be (e.g. in call_centre.views.EligibilityCheckViewSet)

andyhd:
@jenkins-moj retest this please

cla_backend/apps/checker/views.py
```diff
@@ -53,6 +54,9 @@ class EligibilityCheckViewSet(
 ):
     serializer_class = EligibilityCheckSerializer
 
+    def get_request_user(self):
```
marcofucci: you should return the user ;-)

Also, there's already a util func for getting the web user `get_web_user` already imported in this file

andyhd:
@jenkins-moj retest this please

PR #310, opened 2015-03-24
add ciprian

ciprianc:


ciprianc:
auto merging as this branch is completely throw-away. I'm only playing with something.

PR #312, opened 2015-04-07
Feature/sla bugfixes

r4vi:
get_sla_time wasn't working in localtime previously which caused bugs with selecting call backs. This branch changes all uses of get_sla_time to be in the local time zone


andyhd:
:+1:

PR #314, opened 2015-04-14
Feature/scope

s-block:


cla_backend/apps/checker/views.py
```diff
@@ -134,3 +135,25 @@ def post_save(self, obj, created=False):
                 if form.is_valid():
                     form.save(obj.created_by)
                     notify_callback_created(obj)
+
+
+class DiagnosisViewSet(
+    PublicAPIViewSetMixin,
+    DiagnosisModelMixin,
+    mixins.CreateModelMixin,
+    mixins.RetrieveModelMixin,
```
r4vi: does this mean anyone can get back a diagnosis if they know the id? maybe we can lock it based on a session?

r4vi:
another thought is do we want to keep around checker diagnosis objects lying around forever? maybe we can have a clear-up task

s-block:
I was thinking about creating a celery task to delete them after a day or so. The diagnosis is never attached to the eligibility check so could be deleted after the contact form has been submitted

r4vi:
yeah that doesn't need to be part of this PR anyway - more worried about someone brute forcing the guid space (small risk but we don't have any real ratelimit)

r4vi:
:+1:

PR #321, opened 2015-04-17
add UI for triggering a background job that emails the anonomized db dump to

r4vi:
people do will feed it into a management information system.

the email address it uses is in the config repo


andyhd:
Looks ok to me :+1:

PR #322, opened 2015-04-17
the OBIEE zip file will now be encryted with a password stored in the config

r4vi:
repo (different per environment


marcofucci:
:+1:

PR #328, opened 2015-04-27
Feature/scope amends

axemonkey:
* add the 56 day copy change to scope diagnosis
* change the outcome to INELIGIBLE if the problem category is in scope but the problem is not
* bump cla_common to 0.1.72 to account for the new INELIGIBLE outcome


andyhd:
:+1:

PR #354, opened 2015-05-28
MI report export fix for incorrect password

ushkarev:
previously, a server error was sent in response to an incorrect key for encrypted diversity data, now an error message is presented

e.g. see http://sentry.prod.cla.dsd.io:8080/default/default/group/312/

[Fixes #95523210]


stuartornum:
LGTM

PR #360, opened 2015-06-09
CHS: Call Centre case filtering

tyom:
Required for [cla_frontend#553](https://github.com/ministryofjustice/cla_frontend/pull/553)

[PT](https://www.pivotaltracker.com/story/show/91393910)

### DIFF
--- a/cla_backend/apps/call_centre/tests/api/test_case_api.py
+++ b/cla_backend/apps/call_centre/tests/api/test_case_api.py
@@ -655,6 +655,96 @@ class SearchCaseTestCase(BaseSearchCaseAPIMixin, BaseCaseTestCase):
         )
 
 
+class FilteredSearchCaseTestCase(BaseCaseTestCase):
+    def setUp(self):
+        """
+            obj1 created by current operator (phone implicit)
+            obj2 created by another operator (phone implicit)
+            obj3 case by phone (explicit)
+            obj4 case from web (explicit)
+            obj5 case with EOD (web explicit)
+        """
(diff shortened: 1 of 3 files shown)


ushkarev:
:+1:

cla_backend/apps/call_centre/tests/api/test_case_api.py
```diff
@@ -655,6 +655,124 @@ def test_list_with_person_ref_param(self):
...
+            [c['reference'] for c in response.data['results']],
+            ['ref1', 'ref2', 'ref3', 'ref4', 'ref5']
+        )
+
+    def test_phone_cases(self):
+        response = self.client.get(
+            u'%s?only=phone' % self.list_url, format='json',
+            HTTP_AUTHORIZATION=self.get_http_authorization()
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
```
s-block: could create new method to reduce repeated code with something like:

```python
def assertValidResponse(self, response, expected):
    self.assertEqual(response.status_code, status.HTTP_200_OK)
    self.assertEqual(len(expected), response.data['count'])
    self.assertItemsEqual(
        [c['reference'] for c in response.data['results']],
        expected
    )
```

tyom: Good point. [Updated](https://github.com/ministryofjustice/cla_backend/commit/ba71a594814273ab9c1163048c41be0609d91af9).

PR #361, opened 2015-06-10
scope graphs use "permanent" IDs

ushkarev:
so that graph node rearrangement does not affect tests nor URLs (important for google analytics)
[Fixes #95035940]


s-block:
:+1:

PR #363, opened 2015-06-19
reasons for contacting can now be stored in DB

ushkarev:
in public site, when a user drops out of the scope/means test flow, they are presented with a form to find out why they want to get in touch. the form was previously sent to zendesk
[Fixes #96775836]
depends on [cla_common#31](https://github.com/ministryofjustice/cla_common/pull/31)


s-block:
:+1:

PR #366, opened 2015-06-23
new MI report for listing duplicate cases

ushkarev:
based on matching: full name (first name & surname cannot be distinguished), dob & post code
[Fixes #96668304]


s-block:
:+1:

PR #368, opened 2015-06-24
move local settings import so all settings can be overridden

s-block:


ushkarev:
makes sense

marcofucci:
Looks great, I don't know how it ended up there really, it should have always been at the end of the file.

PR #370, opened 2015-06-25
Feature/scope graph tweak

ushkarev:
immigration scope tweaks, new nodes. also, templated scope GraphML files exclude unnecessary elements


s-block:
:+1:

PR #377, opened 2015-07-15
Performance platform tasks

s-block:


ushkarev:
the actual transactions-by-channel calculation aside (i don't know what it's supposed to include), :+1:

PR #383, opened 2015-07-27
Use date in sql for voice lookup, not the full timestamp as timezone …

s-block:
…adjusted


s-block:
retest this please

ushkarev:
:+1:

PR #389, opened 2015-08-05
Feature/csv upload

s-block:


ushkarev:
:+1: (if dates are always exported in the same format)

PR #406, opened 2015-10-02
complaints updates

ushkarev:
• a single case (i.e. its one-to-one EOD) can have multiple open complaints
• cases API reports the open complaint count


s-block:
:+1:

PR #414, opened 2015-11-13
Feature/case extract

s-block:


cla_backend/apps/diagnosis/management/commands/delete_public_diagnosis.py
```diff
@@ -0,0 +1,27 @@
...
+from diagnosis.models import DiagnosisTraversal
+
+
+class Command(NoArgsCommand):
+
+    help = 'Deletes public diagnosis that are more than a day old'
+
+    def handle_noargs(self, *args, **options):
+        self.stdout.write('Total DiagnosisTraversal objects: %s' %
+                    DiagnosisTraversal.objects.all().count())
```
andyhd: Odd indentation here, I think it's better as:
```python
self.stdout.write(
    'Total DiagnosisTraversal objects: {count}'.format(
        count=DiagnosisTraversal.objects.all().count()))
```

cla_backend/apps/diagnosis/management/commands/delete_public_diagnosis.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+import datetime
+from django.core.management.base import NoArgsCommand
```
andyhd: PEP8 - there should be a blank line between core library imports and 3rd party imports

cla_backend/apps/diagnosis/management/commands/delete_public_diagnosis.py
```diff
@@ -0,0 +1,27 @@
...
+        diags = DiagnosisTraversal.objects.filter(
+            case__isnull=True,
+            modified__lte=yesterday,
+        )
+        self.stdout.write('Deleting %s DiagnosisTraversal objects' % diags.count())
+
+        diags.delete()
+
+        self.stdout.write('Total DiagnosisTraversal objects: %s' %
+                    DiagnosisTraversal.objects.all().count())
```
andyhd: `string.format` is better than `%`

cla_backend/apps/reports/forms.py
```diff
@@ -200,6 +200,22 @@ def get_headers(self):
             "Username", "Has_Third_Party", "Time_to_OS_Action"
         ]
 
+    def get_rows(self):
+        for row in self.get_queryset():
+            l = list(row)
+            diversity_json = l.pop() or {}
```
andyhd: I'm guessing this is to get the last item of the row. This isn't obvious enough. Maybe better to use a constant, eg: `diversity_json = l[DIVERSITY_COLUMN]`?

PR #420, opened 2015-12-21
Optimise reports

s-block:


jsugarman:
:+1: looks awesome. i can only dream of such python writing skills

PR #428, opened 2016-01-27
psql will take username as default- could error if no db

EdwardAndress:
- psql will error if there's no db with the same name as the username
- maybe best to have the user connect to one of the template dbs


s-block:
Can one of the admins verify this patch?

s-block:
Can one of the admins verify this patch?

s-block:
add to whitelist

PR #431, opened 2016-02-12
add option to void complaints; exclude from MI

EdwardAndress:


cla_backend/apps/complaints/views.py
```diff
@@ -161,14 +171,15 @@ def add_event(self, request, pk):
     @action()
     def reopen(self, request, pk):
         obj = self.get_object()
-        closed_logs = obj.logs.filter(code='COMPLAINT_CLOSED')
+        closed_logs = obj.logs.filter(code__in=['COMPLAINT_CLOSED', 'COMPLAINT_VOID'])  # complaint void
```
mattpep: this comment appears to be redundant (it's already explained in the code)

cla_backend/apps/complaints/views.py
```diff
@@ -161,14 +171,15 @@ def add_event(self, request, pk):
...
         if not closed_logs.exists():
-            return DRFResponse('Cannot reopen a complaint that is not closed',
+            return DRFResponse('Cannot reopen a complaint that is not closed or void',
```
mattpep: Perhaps I'm grumbling, but I've a slight dislike of double negatives. Could this be re-worded to something like "Only complaints which are closed or voided can be reopened", or perhaps "This complaint is already open".  (Not knowing the states that a case could be in, I cannot say for sure that this is valid assertion)

ibrechin:
:+1:

PR #439, opened 2016-04-08
Add ping.json

s-block:


cla_backend/apps/status/views.py
```diff
@@ -37,6 +39,16 @@ def status(request):
 
 
 @csrf_exempt
+def ping(request):
```
s-block: can be removes as now using pin.json view from moj_irat

cla_backend/urls.py
```diff
@@ -5,6 +5,8 @@
 from django.conf.urls import patterns, include, url
 from django.conf.urls.static import static
 
+from moj_irat.views import HealthcheckView
```
s-block: is this used here?

requirements/base.txt
```diff
@@ -44,3 +44,6 @@ celery==3.1.17
 boto==2.39.0
 PyYAML==3.11
 pyminizip==0.2.1
+
+#Irat healthcheck and ping package
+git+https://github.com/ministryofjustice/django-moj-irat.git
```
s-block: add commit id or version and name

cla_backend/apps/status/views.py
```diff
@@ -1,3 +1,5 @@
+import os
```
s-block: this isn't used so can be removed

PR #440, opened 2016-04-21
Fix Sentry errors

s-block:
- add getter for case ref from eligibility check on puyblic checker api
 - aims to fix error of broken requests from public when saving a case


s-block:
retest this please

jimnarey:
Looks good.

PR #443, opened 2016-06-21
add case ref columns to personal details

s-block:
- related case refs for person related to case
 - related cases for third party persons


cliffxuan:
looks good

PR #452, opened 2018-01-05
Updates copy because of change to legislation.

aliuk2012:
Unfortunately the welsh copy was not translated and therefore it will be in English, There will be
another piece of work in the future that will update all translations from English to Welsh.


cla_backend/apps/diagnosis/data/graph.graphml
```diff
@@ -7571,7 +7571,7 @@ eg divorce, contact with children, finances]]></data>
       </data>
     </node>
     <node id="n314">
-      <data key="d3"><![CDATA[Domestic abuse (during the last 5 years, or their abuser has a current criminal conviction)]]></data>
+      <data key="d3"><![CDATA[Domestic abuse (or their abuser has a criminal conviction)]]></data>
```
misaka: Just to check ... is the use of "or" right here? What's on the left side, "Domestic abuse"?

PR #459, opened 2018-03-05
fixes an issue when trying to install from archives

aliuk2012:
E: Unable to fetch some archives, maybe run apt-get update or try with
--fix-missing?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,23 +23,44 @@ RUN rm -rf /etc/service/sshd /etc/my_init.d/00_regen_ssh_host_keys.sh
 
 # Dependencies
 RUN DEBIAN_FRONTEND='noninteractive' \
-  apt-get -y --force-yes install bash apt-utils python-pip python-dev build-essential git software-properties-common python-software-properties libpq-dev g++ make libpcre3 libpcre3-dev \
-  libxslt-dev libxml2-dev && \
-  apt-get -y build-dep python-psycopg2
+  apt-get update && apt-get install -y \
+    apt-utils \
+    bash \
+    build-essential \
+    g++ \
+    git \
+    libpcre3 \
+    libpcre3-dev \
+    libpq-dev \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN rm -rf /etc/service/sshd /etc/my_init.d/00_regen_ssh_host_keys.sh
 
 # Dependencies
 RUN DEBIAN_FRONTEND='noninteractive' \
-  apt-get -y --force-yes install bash apt-utils python-pip python-dev build-essential git software-properties-common python-software-properties libpq-dev g++ make libpcre3 libpcre3-dev \
+  apt-get update && apt-get -y --force-yes install bash apt-utils python-pip python-dev build-essential git software-properties-common python-software-properties libpq-dev g++ make libpcre3 libpcre3-dev \
```
misaka: A couple things from the [Docker best-practices page](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#minimize-the-number-of-layers):
- combine the packages lists from the two `apt-get install` commands
- for legibility, sort them and put them into a list with one package per line
- clear out the apt cache as part of the same RUN, as each RUN creates a layer (see the [example in CT's Dockerfile](https://github.com/ministryofjustice/correspondence_tool_staff/blob/develop/Dockerfile.base#L13))

Dockerfile
```diff
@@ -23,23 +23,43 @@ RUN rm -rf /etc/service/sshd /etc/my_init.d/00_regen_ssh_host_keys.sh
...
+    libxslt-dev \
+    libxml2-dev \
+    make \
+    python-dev \
+    python-pip \
+    python-software-properties \
+    software-properties-common && \
+  apt-get -y build-dep \
+    python-psycopg2
+
```
misaka: You need to rm the apt cache as I mentioned

PR #470, opened 2018-07-09
LGA-83 Move unit tests and Docker build to Circle CI

rowlando:

- Adds CircleCI job for Python unit tests.
- Adds CircleCI job for building the `cla_backend` Docker image.

As a follow-up, we should remove both of these functions from our self-hosted Jenkins instance.

## Any other changes which would benefit highlighting?

The Docker images are assumed to be stored in Amazon Elastic Container Registry. This might change in the future to Docker Hub or quay.io.

The Python dependencies `psycopg2` and `pyminizip` were updated in order to get the unit tests to pass on Circle CI. The tests run in `circleci/python:2.7` docker image, which differs somewhat from the fatter production Docker image, which contains many more packages. Upgrading `pyminizip` resulted in a broken test because the method signature of `pyminizip.compress_multiple`.

One test is skipped because it is flaky. We'll need a Python developer to fix this test.

### Post-merge steps

- [ ] Disable previous Jenkins job that was building the Docker image
- [ ] Deployment job in Jenkins modified to use the new Docker registry


.circleci/config.yml
```diff
@@ -0,0 +1,81 @@
...
+          command: |
+            docker push $ECR_ENDPOINT/$APP_NAME:$CIRCLE_SHA1
+            docker push $ECR_ENDPOINT/$APP_NAME:$CIRCLE_BRANCH
+
+  test:
+    docker:
+      - image: circleci/python:2.7
+        environment:
+          DB_NAME: circle_test
+          DATABASE_URL: postgresql://root@localhost
```
sldblog: For documentation purposes, it would probably be nice to include the port and database as well, e.g. `postgres://root@localhost:5432/circle_test`.

.circleci/config.yml
```diff
@@ -0,0 +1,81 @@
...
+          DJANGO_SETTINGS_MODULE: cla_backend.settings.circle
+      - image: circleci/postgres:9.4.18-alpine
```
sldblog: _nitpick_ It's probably not necessary to mention the patch version, e.g. we could use `9.4-alpine` as the tag, so we can avoid updating this file too many times.

.circleci/config.yml
```diff
@@ -0,0 +1,81 @@
...
+        environment:
+          POSTGRES_USER: root
+          POSTGRES_DB: circle_test
+    steps:
+      - checkout
+      - run: sudo apt-get install postgresql-client-9.6
```
sldblog: This client version mismatches the PostgreSQL Docker image version. Is that intentional?

.circleci/config.yml
```diff
@@ -0,0 +1,81 @@
...
+      - save_cache:
+          key: pip-v2-{{ checksum "requirements/base.txt" }}-{{ checksum "requirements/testing.txt" }}
+          paths:
+            - "~/.cache/pip"
+
+      - run:
+          name: Run unit tests
+          command: |
+            source env-ci/bin/activate
+            python manage.py test --verbosity=2
```
sldblog: 👏 Nice! TIL

cla_backend/apps/reports/utils.py
```diff
@@ -164,7 +164,7 @@ def generate_zip(self):
         with remember_cwd():
             try:
                 os.chdir(self.tmp_export_path)
-                pyminizip.compress_multiple(glob.glob('*.csv'), self.filename, settings.OBIEE_ZIP_PASSWORD, 9)
+                pyminizip.compress_multiple(glob.glob('*.csv'), [], self.filename, settings.OBIEE_ZIP_PASSWORD, 9)
```
sldblog: I assume this change is due to the package update?

cla_backend/settings/circle.py
```diff
@@ -0,0 +1,26 @@
...
+
+MANAGERS = ADMINS
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'cla_backend.apps.reports.db.backend',
+        'NAME': os.environ.get('DB_NAME', 'circle_test'),
+        'USER': os.environ.get('DB_USER', 'root'),
+        'PASSWORD': os.environ.get('DB_PASSWORD', ''),
+        'HOST': os.environ.get('DB_HOST', 'localhost'),
```
sldblog: I'm curious. We specify the `DATABASE_URL` in the CircleCI configuration (which can have user/password/database sections) but we define the arguments here again. Do we need both?

rowlando: I believe Django uses the keys `NAME`, `USER`, `PASSWORD` and `HOST`.

You're right about then environment variables, we don't necessarily have to specify defaults in this file, but I thought we may as well.

The DATABASE_URL is for convenience only in the Setup Postgres step, but I can see how it would break things if the env variables were changed. I'll scrap the DATABASE_URL and reference the env vars directly in that step.

kerin: @rowlando you can use `DATABASE_URL` via this package though: https://github.com/kennethreitz/dj-database-url

👋

rowlando: Thanks @kerin 

I'll leave it to a Python developer to sort out. Apparently we're getting one soon!

PR #474, opened 2018-08-29
Add Docker build test to validate expected timezone

rowlando:

This pull request tests that the expected timezone is set and sets the timezone to `Europe/London`.

It looks like the timezone has been UTC since 31/01/2018 ([see build job](https://ci.service.dsd.io/view/CLA/job/BUILD-cla_backend/48/console)), which you can see by this screenshot is a result of an error:

timezone-error

## Other, relevant context
It's assumed by looking at the `Dockerfile`, that the timezone was meant to remain as `Europe/London`. It's not obvious what effect this has had on `cla_backend` since the bug made its way into production, as we've had no reported user feedback around this issue.

`cla_public` suffered from a similar issue, which did impact users. You can find out about that in this pull request:

https://github.com/ministryofjustice/cla_public/pull/741

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,81 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: docker:18-git
+    environment:
+      APP_NAME: "get-access/cla_backend"
+    steps:
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run:
+          name: Login to container registry
+          command: |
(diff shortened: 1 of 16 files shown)


sldblog:
📝 The image build now correctly fails on validating the timezone.

sldblog:
@rowlando Could you comment on what is left to make this pull request mergeable, please?

said-moj:
Closing due extremely out of date PR. Only relevant part is the timezone validation which could be raised as a separate PR

PR #477, opened 2018-09-19
LGA-274 Document production release window in README

rowlando:

This documents when a release can be made to production and why there is a release window.

None.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,11 @@ If you get the error `django.db.utils.OperationalError: FATAL:  role "postgres" 
 
 ### Releasing to production
 
+>:warning: Release to production outside of business hours
+>- **Business hours**: 09:00 to 20:00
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -124,6 +124,11 @@ If you get the error `django.db.utils.OperationalError: FATAL:  role "postgres"
 
 ### Releasing to production
 
+>#### :warning: Release to production outside of business hours
+> __Business hours__: 09:00 to 20:00  
+>__Why?__ Any downtime on the frontend and backend between 07:00 and 20:00 - can have serious consequences, leading to shut down of the court legal advice centres, possible press reports and maybe MP questions
```
sldblog: This line mentions 7 am as the period start and the above mentions 9 am. Which one is correct?

PR #478, opened 2018-10-16
Use Cloud Platform's docker registry

rowlando:

This pull request pushes Docker images to Cloud Platform's ECR repo instead.

There is a test change in `cla_backend/apps/legalaid/tests/test_helpers.py:SLATimeHelperTestCase` due to a test failure around daylight savings. The original test failure is in https://circleci.com/gh/ministryofjustice/cla_backend/126.

Please see inline code comments for reasoning. There are two major changes in the test:

- test names reflect the expected outcome (instead of "works")
- day and time mutations were made explicit by the `self._change` function

### DIFF
--- a/cla_backend/apps/legalaid/utils/sla.py
+++ b/cla_backend/apps/legalaid/utils/sla.py
@@ -30,7 +30,7 @@ def get_next_business_day(start_date):
     return filter(lambda x: x.date() > start_date, available_days(365))[0]
 
 
-def get_sla_time(start_time, minutes):
+def get_sla_time(start_time, minutes_delta):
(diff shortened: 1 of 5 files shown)


cla_backend/apps/legalaid/utils/sla.py
```diff
@@ -30,7 +30,7 @@ def get_next_business_day(start_date):
     return filter(lambda x: x.date() > start_date, available_days(365))[0]
 
 
-def get_sla_time(start_time, minutes):
+def get_sla_time(start_time, minutes_delta):
```
rowlando: Much better name and naming is hard. 🥇

PR #481, opened 2018-11-13
Comment out 'flaky' skipped test

Katy600:
This pull request reenables a flaky test that had previously being skipped. It also adds a comment including a reference to the ticket.
While trying to investigate this bug I commented out the skipped test, and ran the test in various scopes, such as: running it alone, in the test class and the in entire test suit. In each instance this test seemed to consistently pass. I then rolled the test back to the commit where it had been failing and tested it in the various scopes there too. The only failure in both cases was a huge number of other unrelated tests failing in another app as a result of outdated lxml.  Once this was upgraded to the latest version these tests were able to pass. 

This test has now been fixed. It was a result of incorrect mocking for the value, 'disposable_capital_assets' which was used inside the method 'is_eligible' called by the EligibilityChecker class. 
Once corrected it passed circleci tests. 

[Ticket for Re-enable skipped Python test in cla_backend](https://dsdmoj.atlassian.net/browse/LGA-84)

## Update 08/01/2019
In response to Patricks next steps, the flaky test was re written with the input of the business analysts to ensure that the test actually reflected the business case. The test was also split into two, so that it was able to test disposable capital separately from the disposable income. This was to ensure that when the is_eligible method was called, which incorporated checking if either of these cases were above the limit, disposable income, which came later in the method, would be reached. 
The doc strings were also updated to explain more clearly the intentions of the test. It now passes both locally and in Circle. It was then refactored to create a dummy citizen that could be reused in each test. 

The mock that was used to fix this was copied from another test. This mock was used repeatedly through out the test suit. This should be refactored at some point but it is currently out of scope.


calv-w:
Hi Katy

I don’t recognise this code.  It’s not anything to do with me!

Thanks & Regards
Wendy

Wendy Calverley | MI developer | LAA Digital, Justice Digital & Technology, 4th Floor, 102 Petty France, London, SW1H 9AJ
Telephone: 07469 579022

sldblog:
Hmm, it seems the service improvement team is part of `laa-get-access` team, which is the owner for this repository. We would have to re-think ownership assignments.

calv-w:
It may be someone on my team, but it’s not me!

Thanks & Regards
Wendy

Wendy Calverley | MI developer | LAA Digital, Justice Digital & Technology, 4th Floor, 102 Petty France, London, SW1H 9AJ
Telephone: 07469 579022

PR #482, opened 2018-11-14
Fix unicode issue in the case handling system

Katy600:

This pull request fixes an error where anything entered into the search bar in the case handling system that was not an ASCII character, such as the fancy apostrophe in Mark O’Brien or when searching with the "£" sign, would result in a 500 error code and display the error message below. 
_UnicodeEncodeError: 'ascii' codec can't encode character u'\u2019' in position 1: ordinal not in range(128)_
This was fixed by simply add 'u' where the search string was being formatted so that it was able to handle  non-ASCII characters such as these. 

Intentionally left blank.

### DIFF
--- a/cla_backend/apps/legalaid/views.py
+++ b/cla_backend/apps/legalaid/views.py
@@ -479,7 +479,7 @@ class FullCaseViewSet(
                 )
             )
             for _ in range(number_of_placeholders):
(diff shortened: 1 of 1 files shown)


Katy600:
Thank you Patrick and David. Yes, I shall add a new ticket to the back log that will add in a test for it. :)

PR #483, opened 2018-11-19
LGA-275: Missing outcome codes

farrepa:

1. Fix Case.outcome_code field:

    - Remove null attribute. Django CharFields should only use empty strings.
    - Migrate data to replace null/None with empty string
    - Update tests
 
2. Update logging. Write to stdout for compatibility with logging infra.
3. Add management command to check for missing outcome codes and log the outcome.
4. Log when fields are denormalised from event Log objects to related Case.
5. Log denormalisation outcome. Test.

Minor changes:

- Clarify repo specific `docs/conf.py`

Deploy repo configures periodic management command invocations


sldblog:
> Remove null attribute. Django CharFields should only use empty strings.

🙋‍♂️ I was curious and found this as reference: https://docs.djangoproject.com/en/1.7/ref/models/fields/#django.db.models.Field.null

> Avoid using `null` on string-based fields such as `CharField` and `TextField` because empty string values will always be stored as empty strings, not as `NULL`. If a string-based field has `null=True`, that means it has two possible values for “no data”: `NULL`, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not `NULL`.

Is this what you are referring to?

cla_backend/settings/docker.py
```diff
@@ -20,30 +21,27 @@
...
+    'filename': '/var/log/wsgi/debug.log',
+    'maxBytes': 1024 * 1024 * 5,  # 5 MB
+    'backupCount': 7,
     'formatter': 'verbose',
     'filters': ['require_debug_true'],
 }
 
 LOGGING['loggers'][''] = {
-    'handlers': ['production_file', 'debug_file'],
+    'handlers': ['production_file', 'debug_file', 'console'],
```
sldblog: 🙋‍♂️ Can we remove both `production_file` and `debug_file` handlers in favour of `console`? If we need to add more granular logging, we could add `LOG_LEVEL` environment variable to switch around.

docs/conf.py
```diff
@@ -21,9 +24,11 @@
 sys.path.insert(0, os.path.abspath('../cla_backend/libs/'))
 sys.path.insert(0, os.path.abspath('..'))
 sys.path.insert(0, os.path.abspath('../../cla_docs/docs/source/'))
-from common_conf import *
+from common_conf import *  # cla_docs.source.common_conf
 
 extensions += ['docs.swag']
+# extensions += ['docs.swag', 'm2r']
+# source_suffix = ['.rst', '.md']
```
sldblog: 🙋‍♂️ Do we need the comments in these two lines?

cla_backend/apps/cla_butler/management/commands/monitor_missing_outcome_codes.py
```diff
@@ -0,0 +1,50 @@
...
+
+from cla_eventlog.constants import LOG_LEVELS, LOG_TYPES
+from cla_eventlog.models import Log
+from legalaid.models import Case
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Count and alert when outcome codes denormalized to Case instances are missing'
```
sldblog: 🙋‍♂️ This description and the surrounding code can be a bit hard to understand if someone is not involved in the missing outcome codes ticket.

Someone might look at logs and see a warning that says `Cases found missing denormalized outcome codes` and might be wondering if that's a problem or not (might be a red herring with live issues).

What do you think about rephrasing to something like "Found high importance cases with mismatching outcome codes; the outcome should have been saved on the case too"?

cla_backend/apps/cla_butler/management/commands/monitor_missing_outcome_codes.py
```diff
@@ -0,0 +1,50 @@
...
+        else:
+            self.stdout.write('Not doing housekeeping because running on secondary instance')
+
+    @staticmethod
+    def check_for_missing_outcome_codes():
+        outcomes_that_should_be_denormed = Log.objects.filter(level=LOG_LEVELS.HIGH, type=LOG_TYPES.OUTCOME)
+        outcomes_missing_denormed_code = outcomes_that_should_be_denormed.filter(case__outcome_code='')
+        cases_missing_denormed_code_pks = outcomes_missing_denormed_code.values_list('case__id', flat=True)
+
+        cases_to_re_denorm = Case.objects.filter(id__in=cases_missing_denormed_code_pks, outcome_code_id__isnull=False,
```
sldblog: _nitpick, depends if we aim this code to be long-lived_
Would we run into query size limits by passing down ids explicitly as a filter (instead of a subquery)?

cla_backend/apps/legalaid/migrations/0020_fill_missing_outcome_codes.py
```diff
@@ -0,0 +1,38 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def re_denormalize_outcome_codes_to_cases(apps, schema_editor):
+    try:
+        from cla_eventlog.constants import LOG_LEVELS, LOG_TYPES
+        from cla_eventlog.models import Log
+    except ImportError:
+        pass
```
sldblog: If this exception happens, would the migration be considered successful? Would there be a way to retry?
Is it safe to do data fixes during migrations?

cla_backend/apps/cla_butler/management/commands/monitor_missing_outcome_codes.py
```diff
@@ -0,0 +1,50 @@
...
+        if self.should_run_housekeeping(**options):
+            self.check_for_missing_outcome_codes()
+        else:
+            self.stdout.write('Not doing housekeeping because running on secondary instance')
+
+    @staticmethod
+    def check_for_missing_outcome_codes():
+        outcomes_that_should_be_denormed = Log.objects.filter(level=LOG_LEVELS.HIGH, type=LOG_TYPES.OUTCOME)
+        outcomes_missing_denormed_code = outcomes_that_should_be_denormed.filter(case__outcome_code='')
+        cases_missing_denormed_code_pks = outcomes_missing_denormed_code.values_list('case__id', flat=True)
```
sldblog: Depending on how/when this command runs, would the above query be efficient or do we need to time it to less busy periods to avoid disruption to the operator's work?

cla_backend/apps/legalaid/migrations/0020_fill_missing_outcome_codes.py
```diff
@@ -0,0 +1,38 @@
...
+    else:
+        outcomes_that_should_be_denormed = Log.objects.filter(level=LOG_LEVELS.HIGH, type=LOG_TYPES.OUTCOME)
+        outcomes_missing_denormed_code = outcomes_that_should_be_denormed.filter(case__outcome_code='')
```
sldblog: ❓Is the filter implicitly using `ORDER BY created`, or in other words, is there any chance the time-ordering of `Log` entries matters?

Ie. if there were two or more HIGH outcome entries, would we overwrite a case outcome to an earlier entry?

cla_backend/apps/legalaid/models.py
```diff
@@ -763,6 +763,24 @@ def is_part_of_split(self):
         """
         return self.from_case or self.split_cases.count() > 0
 
+    def log_denormalized_outcome_fields(self):
+        """
+        For LGA-275 log which fields have been denormalized successfully
+        TODO Replace painfully circuitous reload method when refresh_from_db available in Django 1.8
```
sldblog: 📝 Upgrading to Django 1.8 would help with the broken dependencies, too (one of them requires Django 1.8+)

cla_backend/apps/cla_butler/management/commands/monitor_missing_outcome_codes.py
farrepa: Reworded to reference the ticket, the code's temporary investigative nature, and for clarity.

cla_backend/apps/cla_butler/management/commands/monitor_missing_outcome_codes.py
farrepa: Would be fine as the lifetime high-water mark was <350, but reworked query to join 👍

cla_backend/apps/legalaid/migrations/0020_fill_missing_outcome_codes.py
farrepa: It would be considered successful if the exception occurred. This migration isn't critical and the failure is explicitly caught by the monitoring pieces. Anyhow, migration dependencies updated and historical version of the Log model is now referred to obviating the error handling.

Yes, [common practice](https://docs.djangoproject.com/en/1.7/topics/migrations/#data-migrations)
 in Django.

cla_backend/apps/legalaid/migrations/0020_fill_missing_outcome_codes.py
farrepa: Yes, great spot. Explicitly reversed the queryset ordering to result in the latest HIGH+OUTCOME entry for a case getting written last.

cla_backend/apps/cla_butler/management/commands/monitor_missing_outcome_codes.py
farrepa: It's a small query on smallish amount of data. But reworked for efficiency. Worth logging resource usage and initially running only out of hours at first?

cla_backend/settings/docker.py
farrepa: Sure, happy to. Was minimising changes, but would love to start delicately spring cleaning soon.

docs/conf.py
farrepa: Indeed we don't. They relate to a newly spawned ticket, but can better wait in a stash.

farrepa:
> 🙋‍♂️ I was curious and found this as reference: https://docs.djangoproject.com/en/1.7/ref/models/fields/#django.db.models.Field.null
> Is this what you are referring to?

Yep, exactly that 👍

Great review, thanks. Pushed changes for re-review.

farrepa:
> _question_ I don't know anything about commands/butlers, could you explain how and when does the command code run, please?

Cron jobs configured in https://github.com/ministryofjustice/cla_backend-deploy/blob/master/pillar/prod/data.sls#L38

cla_backend/apps/cla_butler/management/commands/monitor_missing_outcome_codes.py
sldblog: Seeing that the linked cronjob runs at 1 am, I think that is fine. 👍

cla_backend/settings/docker.py
sldblog: I'm happy either way 👍

sldblog:
(needs rebase)

PR #484, opened 2018-11-23
LGA-293 Log every time an outcome Log object is saved

farrepa:

Blindly log each time an outcome Log object is saved.

Intentionally left blank.

### DIFF
--- a/cla_backend/apps/cla_eventlog/models.py
+++ b/cla_backend/apps/cla_eventlog/models.py
@@ -48,6 +48,10 @@ class Log(TimeStampedModel):
 
     def save(self, *args, **kwargs):
         super(Log, self).save(*args, **kwargs)
+        if self.type == LOG_TYPES.OUTCOME:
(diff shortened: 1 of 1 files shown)


cla_backend/apps/cla_eventlog/models.py
```diff
@@ -48,6 +48,10 @@ def __unicode__(self):
 
     def save(self, *args, **kwargs):
         super(Log, self).save(*args, **kwargs)
+        if self.type == LOG_TYPES.OUTCOME:
+            logger.info('LGA-293 Saved outcome code {} (Log id: {}, Case ref:{})'.
+                        format(self.case.outcome_code, self.id, self.case.reference))
```
sldblog: 🙋‍♂️ Would there be _any_ potential occurence of `self.case` being `None` here?

farrepa: It would raise an AttributeError, but `self.case` cannot be None here after the save because it's a required field on the model.

PR #489, opened 2018-12-06
LGA-125 Prevent and remove dupe consecutive outcome codes

farrepa:

1. Prevents duplicate consecutive outcome codes from being logged
2. Add a data migration to remove previous occurrences of duplicate consecutive outcome codes

Management command to remove consecutive duplicate outcome codes. Estimated ~10-15 seconds on production. Output from staging available on JIRA ticket.

No transaction/locking/mutex used so this does not prevent simultaneous-ish writes from different backends occurring. That would be overly invasive at this juncture. The previously introduced monitoring mostly (albeit not fully) covers this and will keep us apprised of further consecutive dupes.

If it remains a significant issue we should have a broader architectural discussion.


cla_backend/apps/cla_eventlog/migrations/0005_remove_same_day_consecutive_outcome_codes.py
```diff
@@ -0,0 +1,75 @@
...
+    # First pass. Gather case/day/code with >1 occurrences
+    cases_with_daily_multiple_events = []
+    grouped_results = groupby(outcome_events, key=lambda i: i)
+    for (result, g) in grouped_results:
+        count = sum(1 for _ in g)
+        if count > 1:
+            result['count'] = count
+            cases_with_daily_multiple_events.append(result)
+
+    # Second pass to reverse chronological scan through all log events for a case on a given day
```
sldblog: This is a two-pass clean-up data migration with two for loops. I find it hard to visualise the desired outcome.

Due to the complexity of this code and because it's going to be kept as a data migration, I would feel more confident if it would be unit tested for

- "here's how bad data looks like" -> "here's what we expect it to look like"
- "here's how good data looks like" -> "we expect no changes"

farrepa: Can do, though the desired outcome is super clearly detailed in the attachment to the [originating ticket](https://dsdmoj.atlassian.net/browse/LGA-125).

Can you confirm you've read through `LGA-125 data migration output on staging.txt`? If still not clear, I'll re-frame it with unit tests.

farrepa: As discussed:

- Data migration moved to a management command to separate from deploy
- Log objects serialised in restorable form

PR #512, opened 2019-01-23
LGA-434 - Alert when cla_backend is close to running out of disk space

said-moj:

Send notification to #cla-alerts slack channel when disk space available is below 2% on any of the cla_backend servers

the disk space health checks have been added to the list of health checks carried out using the moj_irat library. 

Currently the Elastic Load Balancers carry the health checks trigger the relevant cloud watch alarms, these then send a notification to the SNS topics which in turn trigger an AWS Lambda. There are two SNS topics one for low priority and another for high priority notifications.

The current lambdas were sending the alerts to the #high-priority-alerts or #low-priority-alerts slack channel. These were not modified as I am not currently sure where else they are used.

Outside of this PR, a new SNS topic CloudWatchToSlackCLAHealthChecks was created along with a new AWS lamda CloudWatchToSlackCLAHealthChecks which when triggered will send the alert to #cla-alerts

### DIFF
--- a/cla_backend/apps/status/healthchecks.py
+++ b/cla_backend/apps/status/healthchecks.py
@@ -1,0 +1,12 @@
+import os
+
+
+def check_disk():
+    stat = os.statvfs(os.getcwd())
+    available_mb = (stat.f_bavail * stat.f_frsize) / (1024.0 ** 2)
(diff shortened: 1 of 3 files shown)


said-moj:
requires LGA-477 "Lock pip to previous version to enable CLA projects to build and release" to pass checks

PR #513, opened 2019-01-24
Use pip version 18

said-moj:

In CircleCI, the test jobs are currently failing

https://www.python.org/dev/peps/pep-0517/ causes problems with cla_common and pip 19 uses it by default.

Using --no-use-pep517 does not work with pip install -r requirements.txt. The workaround is to downgrade to the previous version of pip.

We can upgrade to the next pip version if cla_common is made compatible with https://www.python.org/dev/peps/pep-0517/.

Same as https://github.com/ministryofjustice/cla_public/pull/800

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -100,6 +100,7 @@ jobs:
           command: |
             pip install virtualenv
             virtualenv env
+            pip install pip==18.1
       - restore_cache:
           keys:
             - pip-v2-{{ checksum "requirements/base.txt" }}-{{ checksum "requirements/testing.txt" }}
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -107,6 +107,7 @@ jobs:
           name: Install dependencies
           command: |
             source env/bin/activate
+            pip install pip==18.1
```
sldblog: *not a blocker*
💅 Semantically, I think this command belongs under "Setup Python environment" rather than "Install dependencies".

said-moj: You're right, I moved it now

PR #559, opened 2019-03-28
LGA-568 Allow same day consecutive outcome codes with differing notes

farrepa:

- Allow same day consecutive outcome codes if notes field differs
- Add ConsecutiveOutcomeCodesTestCase

Intentionally left blank.

## Checklist

- [X] Provided JIRA ticket number in the title


farrepa:
> I think it needs a test case for not allowed when notes (and dates) match (other than the special case of them matching cos notes is blank, as we currently have)
> 
> Could also do with one for non-consecutive matches being allowed, but less bothered about that

Done ✅

PR #564, opened 2019-04-18
LGA-618 - Fix 2018 Discrimination Contracts not applying

said-moj:

Fixes the issue with 2018 Discrimination Contracts not validating due to the matter_type1 field not being provided.

2018 Discrimination rows can have NA as the value for the fixed fee code. These were being incorrectly reported as invalid on UAT

## Checklist

### DIFF
--- a/cla_backend/apps/cla_provider/tests/api/test_csvupload_api.py
+++ b/cla_backend/apps/cla_provider/tests/api/test_csvupload_api.py
@@ -11,6 +11,11 @@ from rest_framework import serializers
 from rest_framework.test import APITestCase
 
 import legalaid.utils.csvupload.validators as v
+from legalaid.utils.csvupload.contracts import (
+    CONTRACT_EIGHTEEN_DISCRIMINATION,
+    contract_2018_category_spec,
+    get_applicable_contract,
+)
 from cla_provider.models import Staff
 from core.tests.mommy_utils import make_recipe
 from core.tests.test_base import SimpleResourceAPIMixin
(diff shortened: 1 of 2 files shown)


sldblog:
🙋‍♂️ `[ x] Provided JIRA ticket` seems to have an extra space

- [x] item (`- [x] item`)
- [ x] item (`- [ x] item`)

cla_backend/apps/cla_provider/tests/api/test_csvupload_api.py
```diff
@@ -389,6 +394,12 @@ def test_validator_valid_2018(self):
         validator = self.get_provider_csv_validator()
         self.assertEqual(len(validator.validate()), 2)
 
+    @override_settings(CONTRACT_2018_ENABLED=True)
+    def test_validator_is_2018_discrimination(self):
+        for matter_type in contract_2018_category_spec["discrimination"]["MATTER_TYPE1"]:
+            if get_applicable_contract(datetime.datetime(2019, 9, 1), matter_type) != CONTRACT_EIGHTEEN_DISCRIMINATION:
+                self.fail("Applicable contract is not 2018 discrimination contract")
```
sldblog: As an outsider, understanding the assertion here is a bit challenging. What does "applicable" mean?

said-moj: Different rates and fees apply to cases depending on what contract applies to them.
The different types of contracts are:

- CONTRACT_THIRTEEN
- CONTRACT_EIGHTEEN
- CONTRACT_EIGHTEEN_DISCRIMINATION

The value of Matter Type 1 field determines which contract applies to the current row.

sldblog: Hmm, it sounds like contracts apply to cases. Is that true? Or is the "provider" the contract holder? (In which case I would except a "provider" word appear somewhere in the test case.)

said-moj: I think in the wider business you are right that the provider is the contract holder. However the language used in some of tickets called the set of rules which a case is evaluated against a 'contract':
"As a Contract Manager I need to know whether a case uploaded within a Provider's csv upload has been completed under the 2018 Education contract " (LGA-608).

This PR and the highlighted fail message is about the case not validating against those rules.

sldblog: It feels there will be opportunities to simplify our models :) Not a blocker.

sldblog: (To clarify further) Would `2018 Education contract` mean a mixture of "provider has a 2018 contract", "education area", "case made while provider was under the 2018 contract"?

said-moj: Yes it has those meanings too. I think it can also mean:

A set of rules that apply to a case to qualify it as a 2018 Education contract case

PR #568, opened 2019-04-26
LGA-610 - Apply 2013 validation rules for eligibility codes for 2018 Education and Discrimination cases

said-moj:

Apply 2013 validation rules for eligibility codes for 2018 Education and Discrimination cases

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/legalaid/utils/csvupload/validators.py
+++ b/cla_backend/apps/legalaid/utils/csvupload/validators.py
@@ -693,9 +693,9 @@ class ProviderCSVValidator(object):
                 return [
                     self._validate_outcome_code,
                     self._validate_fee_code_is_na,
-                    self._validate_eligibility_code_2018,
                     self._validate_signposting_code,
                     self._validate_signposting_code_present_for_outcome_code,
+                    self._validate_eligibility_code_2013,
                 ]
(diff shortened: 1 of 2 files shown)


cla_backend/apps/legalaid/utils/csvupload/validators.py
```diff
@@ -693,9 +693,9 @@ def get_extra_validators_for_applicable_contract(self, applicable_contract):
                 return [
                     self._validate_outcome_code,
                     self._validate_fee_code_is_na,
-                    self._validate_eligibility_code_2018,
                     self._validate_signposting_code,
                     self._validate_signposting_code_present_for_outcome_code,
+                    self._validate_eligibility_code_2013,
```
exonian: Does this need renaming?

Is the context that education and discrimination cases have (and will continue to have) fixed fee rules that are the same as the 2013 ones for all cases, or that they're just not moving over to 2018 yet? My understanding is the former, in which case I think it might be confusing for the validator to be called 2013 when it's used for some types of 2018 case

said-moj: Good point

I've renamed this to `_validate_eligibility_code_against_time_spent`

PR #573, opened 2019-05-13
Shrink Docker image

sldblog:

Switches the Docker image to use Alpine Linux and pre-installed nginx and uwsgi, effectively reducing the image size by ~60%.

```
.../laa-get-access/laa-cla-backend    shrink-image.b65eee4           470MB
.../laa-get-access/laa-cla-backend    master                        1.27GB
```

## Todo

- [x] `connect() to unix:////tmp/backend.sock failed (13: Permission denied) while connecting to upstream` from Nginx -- most likely a side-effect of removing the `www-data` user (fixed in c98650bbc71dc53c0d84b1e222beaee89f9409c5)
- [x] can the integration between nginx-uwsgi-celery be tested in this repository, for example by probing ping/healthcheck? (I have a prototype for this but will open a separate pull request for it.)

## Why do these changes?

The current image is 1.27GB in size. This new version that is based on the official Alpine Linux/Python 2.7 image is around ~480MB big.

The reduced image size will

* speed up builds for this repository (5-6 minutes vs the previous 9-10 minutes).
* speed up running the [end-to-end tests](https://github.com/ministryofjustice/laa-cla-e2e-tests)

* Since the image replaced the `/sbin/my_init` supervision process, `supervisord` was introduced as a replacement. Consequently, the `*.service` files were replaced with a single `supervisord-services.ini`.
* The migration is now run as part of the entrypoint process, every time the container boots up.

## Checklist

- [ ] Provided JIRA ticket number in the title, e.g. "LGA-152: Sample title" **Not applicable.**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ RUN apk add --no-cache \
       py2-pip \
       supervisor \
       tzdata \
-      uwsgi
+      uwsgi-python
 
 # To install pip dependencies
 RUN apk add --no-cache \
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,6 +33,10 @@ RUN mkdir -p /var/log/nginx/cla_backend /var/log/wsgi /var/run/celery /var/log/c
     touch /var/log/wsgi/app.log /var/log/wsgi/debug.log && \
     chmod -R g+s /var/log/wsgi
 
+RUN adduser -D www-data -G www-data && \
+    chown -R www-data:www-data /var/log/uwsgi /var/log/nginx/cla_backend && \
+    chmod -R g+s /var/log/wsgi
+
 WORKDIR /home/app/django
 COPY . .
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,111 +1,47 @@
-#
-# CLA_BACKEND Dockerfile all environments
-#
-# Pull base image.
-FROM phusion/baseimage:0.9.22
+FROM alpine:3.9
 
-# Set correct environment variables.
-ENV HOME /root
+RUN apk add --no-cache \
+      bash \
+      gettext \
+      nginx \
+      py2-pip \
+      supervisor \
+      tzdata \
+      uwsgi
 
-# Use baseimage-docker's init process.
-CMD ["/sbin/my_init"]
+# To install pip dependencies
+RUN apk add --no-cache \
+      build-base \
+      git \
+      libxml2-dev \
+      libxslt-dev \
+      postgresql-dev \
+      python2-dev
 
-# Set timezone
-RUN echo $TZ > /etc/timezone && \
-    apt-get update && apt-get install -y tzdata && \
-    rm /etc/localtime && \
-    ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && \
-    dpkg-reconfigure -f noninteractive tzdata && \
-    apt-get clean
-
-# Remove SSHD
-RUN rm -rf /etc/service/sshd /etc/my_init.d/00_regen_ssh_host_keys.sh
-
-# Dependencies
-RUN DEBIAN_FRONTEND='noninteractive' \
-  apt-get update && apt-get install -y \
-    apt-utils \
-    bash \
-    build-essential \
-    g++ \
-    git \
-    libpcre3 \
-    libpcre3-dev \
-    libpq-dev \
-    libxslt-dev \
-    libxml2-dev \
-    make \
-    python-dev \
-    python-pip \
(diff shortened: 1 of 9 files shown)


Dockerfile
```diff
@@ -1,111 +1,47 @@
...
-RUN mkdir -p /var/log/wsgi && \
-    touch /var/log/wsgi/app.log /var/log/wsgi/debug.log && \
-    chown -R www-data:www-data /var/log/wsgi && \
-    chmod -R g+s /var/log/wsgi
 
-RUN  mkdir -p /var/log/nginx/cla_backend
-ADD ./docker/cla_backend.ini /etc/wsgi/conf.d/cla_backend.ini
-
-# Define mountable directories.
-VOLUME ["/data", "/var/log/nginx", "/var/log/wsgi"]
```
sldblog: 🙋‍♂ I'm unsure if these are still needed, so I removed them.

docker/supervisord-services.ini
```diff
@@ -0,0 +1,16 @@
+[program:nginx]
+command=/usr/sbin/nginx -c /etc/nginx/nginx.conf
+stdout_logfile=/dev/stdout # will only be effective if the program is also logging to stdout
+stderr_logfile=/dev/stderr
+
+[program:uwsgi]
+command=/usr/sbin/uwsgi --ini /home/app/django/docker/cla_backend.ini --die-on-term --log-master
```
sldblog: 🙋‍♂ `--die-on-term` is to enable SIGTERM to kill the uwsgi processes, which is preferable with Docker as stopping containers behave as [follows](https://docs.docker.com/engine/reference/commandline/stop/#extended-description):

> The main process inside the container will receive SIGTERM, and after a grace period, SIGKILL.

sldblog:
End to end tests passed:
```
> laa-cla-e2e-tests@1.0.0 test /app
> protractor
    ✂️ snip ✂️
138 specs, 0 failures, 62 pending specs
Finished in 141.183 seconds

> laa-cla-e2e-tests@1.0.0 test-docker /app
> find test/specs/ -type f -exec nightwatch {} +
    ✂️ snip ✂️
OK. 643  total assertions passed. (3m 31s)
🎉 Tests finished!
```

Dockerfile
said-moj: I thought these were needed to persist the apps and webserver logs

said-moj:
> Furthermore, the www-data user and group were removed, which may not
be a desirable thing.

Is this still the case, it looks like you added it back on https://github.com/ministryofjustice/cla_backend/pull/573/commits/c98650bbc71dc53c0d84b1e222beaee89f9409c5

sldblog:
✅ Successfully deployed and tested on demo environment

sldblog:
> Is this still the case, it looks like you added it back on c98650b

Yes, that is right. I'll remove it from the description.

Dockerfile
sldblog: True, I have no idea how logs are collected from the container

sldblog: Still seems to work:
```
/var/log/nginx$ tail -f demo-backend.cla.dsd.io.access.json
{"timestamp_msec": "1557832907.287", ... }
```
That timestamp is Tuesday, May 14, 2019 12:21:47.287 PM GMT+01:00 DST, so it definitely still gets logged on the Docker host machine. 🤷‍♂

sldblog: Oh... The host machine has an nginx on its own 🤦‍♂

sldblog:
```
$ sudo docker exec -ti cla_backend tail -1 /var/log/nginx/cla_backend/access.log
{ "@timestamp": "2019-05-14T12:26:27+01:00", @fields: { "upstream_addr": "unix:////tmp/backend.sock" } }
```

```
$ tail -1 /var/log/nginx/demo-backend.cla.dsd.io.access.json
{"timestamp_msec": "1557833187.286", "@fields": { "upstream_addr": "127.0.0.1:8000" } }
```

said-moj: Is the flow:
ELB->host-nginx->docker-nginx
or
ELB->docker-nginx

sldblog: Yes, there's a host nginx in the middle and the Kibana instance only sees those logs (even historically).

sldblog: It seems we use [beaver](https://python-beaver.readthedocs.io/en/latest/) to ship the logs to logstash, so it eventually shows up in Kibana.

sldblog: *TL;DR* As far as I can see, the `VOLUME` command is safe to remove because it's unused.

PR #575, opened 2019-05-21
LGA-651: Allow users to upload and download files

sldblog:

Attempts to fix the issue preventing users from uploading and downloading large files.

**Explicitly allow www-data access to /var/tmp/nginx**
In Alpine Linux, the default directory by used by nginx is `/var/tmp/nginx`, compared to Ubuntu's `/var/lib/nginx`.

Currently, users cannot upload and download files due to permission issues which are most likely caused by `/var/tmp/nginx` not having correct permissions.

This change was not picked up when I migrated the image from Ubuntu to Alpine and needs amending.

Alternatively, we could run the nginx process as the `nginx` user, but that would need more changes which I would like to defer now.

**Remove default nginx configuration**
There is a default configuration in Alpine which we do not need and can be removed.

**Collapse layers changing permissions for aufs**
Please see details in 15a59a9.

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,15 @@ RUN apk add --no-cache \
       supervisor \
       tzdata \
       uwsgi-python && \
-    rm /etc/nginx/conf.d/default.conf
+    rm /etc/nginx/conf.d/default.conf && \
+    # Create new user and assign permissions in this layer to allow it to work with aufs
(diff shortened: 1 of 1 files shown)


sldblog:
📝 Testing the following on staging:

- [ ] https://dsdmoj.atlassian.net/browse/LGA-651 (provider uploads)
- [ ] https://dsdmoj.atlassian.net/browse/LGA-650 (report downloads)

PR #576, opened 2019-05-31
LGA-644 - Multiple research preferences options

said-moj:

- Creates a new ContactResearchMethod model
- Adds ManyToMany field to the PersonalDetails model to create a relationship to the ContactResearchMethod model
- Creates a migration that creates default contact research methods
- Creates a migration to migrate data from previous contact_for_research_via field to the new field.

Call centre operators would like to select more than one contact method when contacting the user for research.

https://github.com/ministryofjustice/cla_backend/pull/577 must be merged before this PR
## Checklist


cla_backend/apps/legalaid/migrations/0021_auto_20190515_1042.py
```diff
@@ -0,0 +1,53 @@
...
+                ),
+                ("reference", uuidfield.fields.UUIDField(unique=True, max_length=32, editable=False, blank=True)),
+            ],
+            options={"abstract": False},
+            bases=(core.cloning.CloneModelMixin, models.Model),
+        ),
+        migrations.AddField(
+            model_name="personaldetails",
+            name="contact_for_research_methods",
+            field=models.ManyToManyField(to="legalaid.ContactResearchMethod", null=True),
```
sldblog: _curious_
Is this going to be an array field?

said-moj: @sldblog Yep it's moving from a single value for contact preference to multiple values

sldblog: Sorry, what I meant is that I'm curious how this column type looks in PostgreSQL.

sldblog: _not a blocker!_

exonian: m2m through table

cla_backend/apps/cla_provider/serializers.py
```diff
@@ -283,6 +284,10 @@ class Meta(AdaptationDetailsSerializerBase.Meta):
         )
 
 
+class ContactResearchMethodSerializer(ContactResearchMethodSerializerBase):
```
exonian: What is the purpose of not just using the `Base` class? (and then renaming it to remove the `Base` from the name if it's always used directly and never subclassed)

said-moj: Good point. I think I was just copying the pattern for having a separate serialiser for the providers but the fields are the same as the call_centre fields so don't need this empty class. Will remove and update.

PR #577, opened 2019-06-24
Add --noinput flag to migrations so that it doesn't hang during deployment

said-moj:

Add --noinput flag to migrations so that it doesn't hang during deployment

Migrations are run as part of the container entrpoint commans. Without the --noinput flag the migrate command will hang waiting for input, this will cause the deployment to fail as the container does not come up successfully.

## Checklist

- [] Provided JIRA ticket number in the title, e.g. "LGA-152: Sample title"

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -10,7 +10,7 @@ migrations() {
     if [ "$BACKEND_ENABLED" == "True" ]; then
 
         python manage.py install_postgres_extensions
-        python manage.py migrate
+        python manage.py migrate --noinput
 
         python manage.py loaddata initial_groups.json
         python manage.py loaddata initial_category.json


docker/entrypoint.sh
```diff
@@ -10,7 +10,7 @@ migrations() {
     if [ "$BACKEND_ENABLED" == "True" ]; then
 
         python manage.py install_postgres_extensions
-        python manage.py migrate
+        python manage.py migrate --noinput
```
sldblog: Ah, TIL<sup>[source](https://docs.djangoproject.com/en/1.9/ref/django-admin/#django-admin-migrate)</sup> that

> **Changed in Django 1.9:**
> The `--no-input` alias was added.

sldblog:
Does the checklist item in the description apply?

said-moj:
> Does the checklist item in the description apply?

No, should it removed?

PR #579, opened 2019-07-26
LGA-716 - Associate operators with an organisation

said-moj:

Creates new Organisation model

- Added organisation foreign key to Operator model
- Added django command to auto create Agilisys organisation and assign existing operators to organisation
- Restricted organisation operator managers to only seeing operators in their organisation
- Automatically add newly created operator to the same organisation as the operator manager that is creating them
- Added command to assign CLA Superusers 'Can add organisation' and 'Can change organisation' permissions

Associate operators with an organisation so that content can be restricted to a specific organisation

## Checklist

### DIFF
--- a/cla_backend/apps/call_centre/admin/__init__.py
+++ b/cla_backend/apps/call_centre/admin/__init__.py
@@ -1,5 +1,6 @@
 from django.contrib import admin
 from django.core.exceptions import ObjectDoesNotExist
+from django.db.models import Q
 
 from core.admin.modeladmin import OneToOneUserAdmin
 from .forms import OperatorAdminForm, FullOperatorAdminForm, CaseworkerAdminForm
@@ -27,6 +28,22 @@ class OperatorAdmin(OneToOneUserAdmin):
     search_fields = ["user__username", "user__first_name", "user__last_name", "user__email"]
     list_filter = ["organisation__name"]
 
+    def get_list_filter(self, request):
+        return self.list_filter if self._is_loggedin_superuser(request) else []
+
+    def formfield_for_foreignkey(self, db_field, request, **kwargs):
+        if db_field.name == "organisation" and not self._is_loggedin_superuser(request):
+            try:
+                operator = request.user.operator
+            except ObjectDoesNotExist:
+                pass
+            else:
+                if operator.organisation:
+                    kwargs["queryset"] = Organisation.objects.filter(pk=operator.organisation.id)
+                    kwargs["required"] = True
+                    kwargs["initial"] = operator.organisation.id
+        return super(OperatorAdmin, self).formfield_for_foreignkey(db_field, request, **kwargs)
+
     def _is_loggedin_superuser(self, request):
         user = request.user
         if user.is_superuser:
@@ -46,6 +63,20 @@ class OperatorAdmin(OneToOneUserAdmin):
             self.form = self.simple_op_form
 
         return super(OperatorAdmin, self).get_form(request, obj=obj, **kwargs)
+
+    def get_queryset(self, request):
+        qs = super(OperatorAdmin, self).get_queryset(request)
+        if not self._is_loggedin_superuser(request):
+            try:
+                operator = request.user.operator
+            except ObjectDoesNotExist:
+                pass
+            else:
+                query = Q(organisation__isnull=True)
+                query.add(Q(organisation=operator.organisation), Q.OR)
+                qs = qs.filter(query)
+
+        return qs
 
     def has_change_permission(self, request, obj=None):
         """
--- a/cla_backend/apps/call_centre/admin/__init__.py
+++ b/cla_backend/apps/call_centre/admin/__init__.py
@@ -10,6 +10,10 @@ class OperatorAdmin(OneToOneUserAdmin):
     actions = None
     simple_op_form = OperatorAdminForm
     full_op_form = FullOperatorAdminForm
+
+    def operator_organisation(obj):
+        return obj.organisation if obj.organisation else ""
+
     list_display = (
         "username_display",
         "email_display",
@@ -18,8 +22,10 @@ class OperatorAdmin(OneToOneUserAdmin):
         "is_active_display",
         "is_manager",
         "is_cla_superuser",
+        operator_organisation,
     )
     search_fields = ["user__username", "user__first_name", "user__last_name", "user__email"]
+    list_filter = ["organisation__name"]
 
     def _is_loggedin_superuser(self, request):
         user = request.user
--- a/cla_backend/apps/call_centre/admin/__init__.py
+++ b/cla_backend/apps/call_centre/admin/__init__.py
@@ -3,7 +3,7 @@ from django.core.exceptions import ObjectDoesNotExist
 
 from core.admin.modeladmin import OneToOneUserAdmin
 from .forms import OperatorAdminForm, FullOperatorAdminForm, CaseworkerAdminForm
-from ..models import Operator, Caseworker
+from ..models import Operator, Caseworker, Organisation
 
 
 class OperatorAdmin(OneToOneUserAdmin):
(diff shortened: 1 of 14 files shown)


cla_backend/apps/call_centre/admin/__init__.py
```diff
@@ -1,15 +1,20 @@
...
+from ..models import Operator, Caseworker, Organisation
 
 
 class OperatorAdmin(OneToOneUserAdmin):
     actions = None
     simple_op_form = OperatorAdminForm
     full_op_form = FullOperatorAdminForm
+
+    def operator_organisation(obj):
+        return obj.organisation if obj.organisation else ""
```
exonian: `return obj.organisation or ""` is more succinct and I think more readable

cla_backend/apps/call_centre/admin/forms.py
```diff
@@ -1,13 +1,22 @@
 from django import forms
-
```
exonian: I'd prefer to keep this space between the third-party package block and the local package block

cla_backend/apps/call_centre/management/commands/cla_superuser_organisation_perms.py
```diff
@@ -0,0 +1,19 @@
+from django.core.management import BaseCommand
+from django.contrib.auth.models import Group, Permission
+
+
+class Command(BaseCommand):
+    help = "Assign CLA Superusers 'Can add organisation' and 'Can change organisation' permissions"
+
+    def handle(self, *args, **options):
+        group = Group.objects.get(name="CLA Superusers")
+        if not group:
```
exonian: This `.get()` will throw a `Group.DoesNotExist` exception (which is the specific subclass of `ObjectDoesNotExist`) rather than returning something falsey, so you should try/except that

(or use a `filter()` first, check `.exists()` on the queryset that that returns, and then use a `.get()` with no arguments to go from the queryset to the object)

cla_backend/apps/call_centre/management/commands/cla_superuser_organisation_perms.py
```diff
@@ -0,0 +1,19 @@
...
+            return
+
+        can_add_organisation_perm = Permission.objects.get(name="Can add organisation")
+        if can_add_organisation_perm:
```
exonian: As above with regard to use of `.get()`

In this case, though, it might be preferable to just use a filter and pass the result (whether there's anything in it or not) to the `.add()` without checking.

Something like -- _and I haven't tested this_ -- ```group.permissions.add(Permission.objects.filter(name="Can add organisation"))```

Or even combining the two steps
```group.permissions.add(Permission.objects.filter(name__in=["Can add organisation", "Can change organisation"]))```
again, not tested at all :)

File: cla_backend/apps/call_centre/management/commands/create_and_assign_agilisys_organisation.py [hunk 5]
```diff
@@ -0,0 +1,16 @@
+from django.core.management import BaseCommand
+from call_centre.models import Organisation, Operator
+
+
+class Command(BaseCommand):
+    help = "Create Agilisys organisation and assign current agilisys operators to that organisation"
+
+    def handle(self, *args, **options):
+        (organisation, created) = Organisation.objects.get_or_create(name="Agilisys")
```
exonian: This is fine and valid, but I'm not used to seeing the left hand side written with the parentheses
More normal to write `organisation, created = ...` I think

cla_backend/apps/call_centre/management/commands/create_and_assign_agilisys_organisation.py
```diff
@@ -0,0 +1,16 @@
...
+        operators = Operator.objects.filter(organisation__isnull=True, user__email__endswith="@agilisys.co.uk")
+        if not operators.count():
```
exonian: `.exists()` is probably slightly more efficient, so we should use that (or use `.count()` but store it in a var and use it for both the check and the string formatting

exonian: Bit weird to me to have this in the negative. Think it'd read easier as `if` than `if not` (with the stanzas obviously reversed)

cla_backend/apps/call_centre/models.py
```diff
@@ -25,8 +25,16 @@ def save(self, *args, **kwargs):
         return obj
 
 
+class Organisation(TimeStampedModel):
+    name = models.CharField(max_length=255, blank=False, null=False, unique=True)
```
exonian: I wouldn't include the `blank=False` and `null=False` to be honest. They're the defaults and, while it's not wrong to be explicit about them, at a glance one might assume they're being set to non-default values if they're being set (if I weren't code reviewing I think I might have made that mistake)

cla_backend/apps/call_centre/tests/admin/test_views.py
```diff
@@ -10,17 +10,26 @@ class OperatorAdminViewTestCase(TestCase):
...
                 "call_centre.operator",
                 user__username=username,
                 is_manager=is_manager,
                 is_cla_superuser=is_cla_superuser,
+                **kwargs
             )
 
         self.django_admin = make_user(is_staff=True, is_superuser=True)
+        self.foo_organisation = make_recipe("call_centre.organisation", name="Foo org")
+        self.foo_organisation.save()
```
exonian: What does `.save()` achieve here? Is there a post-save action that we need to make sure is kicked off? If not, `make_recipe()` persists the object (whereas I believe it's `prepare_recipe()` that doesn't)

This will apply in a number of other places too, if I'm right

cla_backend/apps/call_centre/views.py
```diff
@@ -484,9 +485,25 @@ class UserViewSet(CallCentrePermissionsViewSetMixin, BaseUserViewSet):
...
+        return qs
+
     def get_logged_in_user_model(self):
         return self.request.user.operator
 
+    @transaction.atomic
+    def create(self, request, *args, **kwargs):
+        operator = self.get_logged_in_user_model()
+        if operator.organisation:
+            request.DATA["organisation"] = operator.organisation.id
```
exonian: Does this have test coverage / has it been fully tested manually?

If you'd made me guess, I'd have said that `DATA` there might have been immutable, but as long as the behaviour's been tested I'm more than happy to concede my guess is wrong.

cla_backend/apps/call_centre/management/commands/cla_superuser_organisation_perms.py
said-moj: `group.permissions.add(Permission.objects.filter(name__in=["Can add organisation", "Can change organisation"]))`

Didn't work for me as group.permissions.add wasn't expecting queryset.

However 
`permissions = Permission.objects.filter(name__in=["Can add organisation", "Can change organisation"])
groups.first().permissions.add(*permissions)
`
has worked for me

cla_backend/apps/call_centre/tests/admin/test_views.py
said-moj: I didn't know that make_recipe was persisting the object.

Amended

cla_backend/apps/call_centre/views.py
said-moj: I've added the missing test coverage `call_centre.tests.api.test_user_api.UserTestCase.test_operator_manager_can_create_operator_of_same_organisation`

>DATA there might have been immutable

I think it was created with `mutable=True` https://docs.djangoproject.com/en/1.7/ref/request-response/#django.http.QueryDict.__init__

cla_backend/apps/call_centre/management/commands/cla_superuser_organisation_perms.py
```diff
@@ -0,0 +1,13 @@
...
+
+class Command(BaseCommand):
+    help = "Assign CLA Superusers 'Can add organisation' and 'Can change organisation' permissions"
+
+    def handle(self, *args, **options):
+        groups = Group.objects.filter(name="CLA Superusers")
+        if not groups.exists():
+            return
+        permissions = Permission.objects.filter(name__in=["Can add organisation", "Can change organisation"])
+        groups.first().permissions.add(*permissions)
```
exonian: Superb. This is exactly what I tried and failed to write, excellent fix

PR #581, opened 2019-08-05
LGA-716 - Associate operators with an organisation

said-moj:

Creates new Organisation model

- Added organisation foreign key to Operator model
- Added django command to auto create Agilisys organisation and assign existing operators to organisation
- Restricted organisation operator managers to only seeing operators in their organisation
- Automatically add newly created operator to the same organisation as the operator manager that is creating them
- Added command to assign CLA Superusers 'Can add organisation' and 'Can change organisation' permissions

- **Added organisation sorting in the admin area**
- **Added unassigned filter for organisation in the admin area**

Associate operators with an organisation so that content can be restricted to a specific organisation

The majority of this work was reviewed in https://github.com/ministryofjustice/cla_backend/pull/579

## Checklist

### DIFF
--- a/cla_backend/apps/call_centre/admin/__init__.py
+++ b/cla_backend/apps/call_centre/admin/__init__.py
@@ -12,10 +12,7 @@ class OrganisationListFilter(admin.SimpleListFilter):
     parameter_name = "organisation"
 
     def lookups(self, request, model_admin):
-        organisations = []
-        for organisation in Organisation.objects.all():
-            organisations.append((organisation.id, organisation.name))
-
+        organisations = list(Organisation.objects.values_list("id", "name"))
         if organisations:
             organisations.append(("", "Unassigned"))
         return organisations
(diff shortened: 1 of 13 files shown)


cla_backend/apps/call_centre/admin/__init__.py
```diff
@@ -7,14 +7,34 @@
 from ..models import Operator, Caseworker, Organisation
 
 
+class OrganisationListFilter(admin.SimpleListFilter):
+    title = "Organisation"
+    parameter_name = "organisation"
+
+    def lookups(self, request, model_admin):
```
exonian: I remembered what I was trying to describe earlier!
It's `queryset.values()`

So you should be able to replace the construction of the organisations list with `Organisation.objects.values('id', 'name')`
You'll need to convert that to a list (`oraganisations = list(organisations)`) before appending to it, I think

cla_backend/apps/call_centre/admin/__init__.py
```diff
@@ -7,14 +7,34 @@
...
+        return organisations
+
+    def queryset(self, request, queryset):
+        value = self.value()
+        kwargs = {}
+        if value:
+            kwargs["organisation__id"] = value
+        elif value == "":
+            kwargs["organisation__isnull"] = True
+
```
exonian: It would be good to have a comment explaining that the other case is value being `None` and why and how that is implicitly handled

PR #585, opened 2019-08-20
LGA-715 - Assign organisation to cases

said-moj:

Added new Case.organisation field
Added script to assign Case.organisation to Operator.Organisation
Assign cases organisation when a case is created

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/legalaid/management/commands/assign_existing_cases_to_organisation.py
+++ b/cla_backend/apps/legalaid/management/commands/assign_existing_cases_to_organisation.py
@@ -8,8 +8,7 @@ class Command(BaseCommand):
 
     def handle(self, *args, **options):
         default_organisation = Organisation.objects.get(name="Agilisys")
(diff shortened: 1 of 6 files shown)


cla_backend/apps/legalaid/management/commands/assign_existing_cases_to_organisation.py
```diff
@@ -0,0 +1,29 @@
...
+            return
+
+        self.stdout.write("Assigning organisations based on creator organisation")
+        for organisation in organisations:
+            qs = cases_qs
+            if organisation:
+                qs = qs.filter(created_by__operator__organisation=organisation)
+            else:
+                organisation = default_organisation.id
+            qs.update(organisation=organisation)
```
exonian: I'm pretty sure I understand all this, but the handling of `None` is a bit difficult to grok. I feel like it'd be neater outside of the organisation loop. I think it'd just be a one-liner, be clearer to read, and would actually make for less code rather than more

PR #590, opened 2019-08-29
Revert "LGA-715 - Assign organisation to cases"

said-moj:
Reverts ministryofjustice/cla_backend#585

Reverting this as it broke production
https://sentry.service.dsd.io/mojds/prod-backend/issues/37309/


sldblog:
Fails test

PR #598, opened 2019-10-23
LGA-793 - Case / Complaints views activity report

said-moj:

Added new AuditLog Model to track case and complaint views
Added auditlog ManyToManyField to Complaint model
Added auditlog ManyToManyField to Case model

## What outcome are we trying to achieve for our users?
We need to keep track who is viewing which cases as there are now multiple organisations using the system and produce a report of such activity for power users.

## Checklist


sldblog:
> ### Any other changes that would benefit highlighting?
> We need to keep track who is viewing which cases as there are now multiple organisations using the system and produce a report of such activity for power users.

_unrelated to the contents of the pull request_

😍 the context! I'm wondering if we should change the leading question to always end up at this kind of answer.

What do you think about changing the pull request template (in another pull request) to ask "What outcome are we trying to achieve for our users?"

Something based on https://youtu.be/2zYxWEZ0gYg?t=1295

.circleci/config.yml
```diff
@@ -137,7 +137,7 @@ jobs:
           name: Run tests with coverage
           command: |
             source env/bin/activate
-            coverage run --source cla_backend manage.py test --verbosity=2
+            coverage run --source cla_backend --omit='*migrations*' manage.py  test --verbosity=2
```
sldblog: 🙋‍♂ Might be worth putting these ignores and the default values to [configuration files](https://coverage.readthedocs.io/en/v4.5.x/config.html#config), so locally we can reproduce the same results without repeating these exact arguments.

said-moj:
>What do you think about changing the pull request template (in another pull request) to ask "What outcome are we trying to achieve for our users?"

I think that's a good a question. I think we should add to existing questions.  I like to keep `Any other changes that would benefit highlighting?` because that's where I can put related PRs.

PR #599, opened 2019-10-29
LGA-850 - Make date range check work when clocks go back

exonian:
# What does this pull request do?
Fixes the date range check on the reports generation form when the date range crosses a date on which the clocks go back.
Adds some solid tests which always test the cases when the clocks go back or forward, rather than only testing the current dates, and which actually test the edges of the allowed date ranges.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/reports/tests/test_reports.py
+++ b/cla_backend/apps/reports/tests/test_reports.py
@@ -107,32 +107,36 @@ class ReportsDateRangeValidationWorks(TestCase):
         class T(reports.forms.DateRangeReportForm):
             max_date_range = 5
 
-        start = datetime.datetime(2019, 3, 29, 12)
-        i = T(data={"date_from": start, "date_to": start + datetime.timedelta(days=4)})
+        start = datetime.datetime(2019, 3, 29, 12)  # GMT
+        end = datetime.datetime(2019, 4, 2, 12)  # BST
+        i = T(data={"date_from": start, "date_to": end})
         self.assertTrue(i.is_valid())
 
     def test_valid_date_range_clocks_going_back(self):
         class T(reports.forms.DateRangeReportForm):
             max_date_range = 5
 
-        start = datetime.datetime(2019, 10, 25, 12)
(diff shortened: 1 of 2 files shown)


cla_backend/apps/reports/tests/test_reports.py
```diff
@@ -76,18 +76,64 @@ def test_mi_survey_dom1_extract_organisation_column(self):
...
+        start = datetime.datetime(2019, 3, 29, 12)
+        i = T(data={"date_from": start, "date_to": start + datetime.timedelta(days=5)})
+        self.assertFalse(i.is_valid())
+
+    def test_invalid_date_range_clocks_going_back(self):
+        class T(reports.forms.DateRangeReportForm):
+            max_date_range = 5
+
+        start = datetime.datetime(2019, 10, 25, 12)
+        i = T(data={"date_from": start, "date_to": start + datetime.timedelta(days=5)})
```
sldblog: I read this test as validating `max_date_range = 5` in relation to `2019, 10, 25, 12 + timedelta(days=5)` is invalid.

I'm afraid I don't understand what this means. Why is this invalid?

sldblog: (What probably confuses me is that we're calculating the `date_to` with `.timedelta` and I don't understand why.)

exonian: It's cos the date range is inclusive, so today to tomorrow is a 2-day range, today to (today + 2) is a 3-day range, and so on.

exonian: > (What probably confuses me is that we're calculating the `date_to` with `.timedelta` and I don't understand why.)

Just a hangover from the tests which use `now()` as the start, rather than a fixed date

sldblog: What do you think about changing the test to use fixed dates?
`"date_from": 2019-10-25 12:00 BST, "date_to": 2019-10-30 12:00 GMT` would make it easier for me to understand what's being asserted here.

exonian: Updated

PR #619, opened 2020-03-19
LGA-1040 Helm and LGA-997 non-root container

exonian:
- Adds a Helm chart with templated resources and variables
- Adds base variables file and env-specific variable files
- Makes the app run as a non-root user (including hugely simplifying what it's trying to do; not compatible with Template Deploy as is)
- Deploys a temporary database pod, according to environment
- Adds Jobs with hooks to make them run automatically during chart install (first deploy) and/or upgrade (subsequent deploys)
- Adds very simple bash scripts to make local and staging deploys extremely simple to run

Intentionally left blank.

## Checklist


said-moj:
Are we OK with this branch being long lived and only going into master momentarily before the cut off? I say this because the changes to the Dockerfile and docker/* files are not compatible with the current deployment

exonian:
> Are we OK with this branch being long lived and only going into master momentarily before the cut off? I say this because the changes to the Dockerfile and docker/* files are not compatible with the current deployment

Yeah I think that's unavoidable tbh. I can't think of a way that supports both, outside of completely separate copies of all the docker stuff, to build and push two separate images, which then doesn't seem any better than not being able to merge a branch. But if you have ideas to make it work they'd be gratefully received cos not being able to integrate doesn't feel good, no

PR #620, opened 2020-03-24
LGA-1046 - Reset the task last message at the start of task.run

said-moj:

Reset the task last message at the start of task.run
Added `Check passphrase and try again` message when diversity passphrase is incorrect

Currently the OBIEE extracts seem to failing with the error message `An error occurred creating the zip file` however the extract seems to be downloadable.

https://docs.celeryproject.org/en/3.0/userguide/tasks.html#instantiation states that:
> A task is not instantiated for every request, but is registered in the task registry as a global instance.

So this means that the same instance of the OBIEEExportTask is being used to handle multiple exports which means unless explicitly reset instance properties such as `message` retain their values through multiple exports.

The current issue is cause by when one export fails due to a legitimate error such as invalid passphrase, all subsequent successful exports will have the same message because `self.message` is only set when there is an error.

## Checklist

### DIFF
--- a/cla_backend/apps/reports/tests/test_tasks.py
+++ b/cla_backend/apps/reports/tests/test_tasks.py
@@ -1,0 +1,42 @@
+import mock
+import datetime
+import json
+from psycopg2 import InternalError
+from django.test import TestCase
+from reports.tasks import OBIEEExportTask
+from core.tests.mommy_utils import make_user, make_recipe
+from legalaid.utils.diversity import save_diversity_data
+
+
+class OBIEEExportTaskTestCase(TestCase):
+    def setUp(self):
+        self.personal_details = make_recipe("legalaid.personal_details")
+        save_diversity_data(self.personal_details.pk, {"test": "test"})
+
+        self.task = OBIEEExportTask()
+        self.task._create_export = mock.MagicMock()
+        self.task.send_to_s3 = mock.MagicMock()
+        self.user = make_user()
+
+    def run_task(self, passphrase):
+        today = datetime.datetime.today()
+        filename = "cla.database.zip"
+        post_data = {"action": "Export", "date_month": today.month, "date_year": today.year, "passphrase": passphrase}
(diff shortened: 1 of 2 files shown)


exonian:
That coverage... 😭

cla_backend/apps/reports/tests/test_tasks.py
```diff
@@ -0,0 +1,40 @@
...
+        with self.assertRaisesMessage(InternalError, "Wrong key or corrupt data"):
+            self.run_task("badpass")
+        self.assertEqual("Check passphrase and try again", self.task.message)
+
+    def test_message_reset_per_run(self):
+        with self.assertRaisesMessage(InternalError, "Wrong key or corrupt data"):
+            self.run_task("badpass")
+
+        self.assertEqual("Check passphrase and try again", self.task.message)
+        self.run_task("cla")
```
exonian: Where is this correct password defined?

said-moj: Update method with the following now:
`        # Run task with an invalid passphrase
        with self.assertRaisesMessage(InternalError, "Wrong key or corrupt data"):
            self.run_task("badpass")
        self.assertEqual("Check passphrase and try again", self.task.message)

        # Run task with the correct passphrase
        self.run_task("cla")
        self.assertEqual("", self.task.message)`

Correct passphrase for testing is "cla"

exonian: I guessed that but I can't work out *why* it's correct

said-moj: The diversity data is encrypted using public/private key mechanism, the development keys are stored in https://github.com/ministryofjustice/cla_backend/tree/master/cla_backend/apps/legalaid/utils/keys (provided as environment variables in other environments)

The private key requires a passphrase which for development is cla

exonian: Ah super, thanks

PR #633, opened 2020-05-19
LGA-999 queue worker sqs

said-moj:

- Adding celery workers deployment
- Updated environment variables to a dict instead of a list so that values from different config files can be merged

### Celery / Kombu patch
Additional changes were required to make the app work with limited permissions that the IAM sqs user has now on the cloud platform environments,  this includes making sure the app is not doing operations such as ListingQueues.
It was not possible to achieve these changes with the core celery/kombu packages. Instead we had to monkey patch kombu.transport.SQS.Transport to use our CLASQSChannel class which noops overreaching actions such as listing of queues.

### Boto
The eu-west-2 region only supports Signature Version 4, the previous signature was generated using version https://docs.aws.amazon.com/general/latest/gr/s3.html.
This was done by setting the environment variables S3_USE_SIGV4 and AWS_S3_HOST

The list of regions that boto knows about is hard coded in a json file that is included in the package. It can read additional regions from a file that is referenced in BOTO_ENDPOINTS environment variable, which is what we have done to get it to use the eu-west-2 region for sqs 

Intentionally left blank.

## Checklist


cla_backend/settings/base.py
```diff
@@ -333,10 +336,22 @@
 
 BROKER_TRANSPORT_OPTIONS = {
     "polling_interval": 10,
-    "region": "eu-west-1",
+    "region": os.environ.get("SQS_REGION", "eu-west-1"),
     "wait_time_seconds": 20,
     "queue_name_prefix": _queue_prefix % {"env": CLA_ENV},
 }
+
+if "CELERY_PREDEFINED_QUEUE_URL" in os.environ:
```
exonian, line 344: I think I would prefer this to check the truthiness of that value, rather than just the existence of the key. I'm just a bit concerned that a value of `None` would break things in a way that you wouldn't expect without reading and understanding this bit of code.

helm_deploy/cla-backend/values-dev.yaml
```diff
@@ -16,5 +16,46 @@ service:
...
+    secret:
+      optional: true
+  DB_PASSWORD:
+    secret:
+      optional: true
+  LOAD_TEST_DATA:
+    value: "True"
+  ADMIN_USER:
+    value: "admin"
+    secret: null
```
exonian, line 37: What are the `secret: null` entries for? Are they to override the defaults from `values.yaml`, and it merges objects so you have to explicitly nullify keys rather than just omitting them?

said-moj: Yes it's to override the defaults from values.yaml and set development values for local development instead of reading them from secrets.

>so you have to explicitly nullify keys rather than just omitting them?
Yes

said-moj: i think it only overrides the sub keys that you specify and not the the entire ADMIN_USER object

cla_backend/settings/base.py
said-moj: updated with `if os.environ.get("CELERY_PREDEFINED_QUEUE_URL"):`

PR #634, opened 2020-05-22
LGA-1053 - Update cla_common for new operator hours

exonian:

This version of cla_common has a new OPERATOR_HOURS constant, which
is used for determining valid callback slots and in SLA calculation

This change affected the tests in a couple of ways.

The smaller one was a test which treated Saturday as a short working day, now updated for it being a non-working day.

The bigger one was around the SLA_480 (8 hour) window. All these tests use a 10am callback time which meant the SLA_480 fell in the evening of the same day, so the tests were checking for it being 8 hours on from the callback time. With the shorter operator hours, this SLA now falls on the next working day instead. That it is exactly 1 day and 0 hours is a characteristic of the new operator hours being 8 hours long also.

With the SLAs now spilling over onto the next day, these tests also had to start using a `mock` for `current_date_time` from `cla_common`, as other of our tests already do (so that the `next_working_day` is correctly calculated for these dates in the past)

## Checklist


cla_backend/apps/call_centre/tests/api/test_case_api.py
```diff
@@ -855,7 +855,7 @@ def _default_local_dt_sla_120(self):
 
     @property
     def _default_local_dt_sla_480(self):
-        return self._default_local_dt + datetime.timedelta(minutes=480)
+        return self._default_local_dt + datetime.timedelta(days=1)
```
said-moj, line 858: There are quite a few changes like these, what's the reasoning behind the timedelta change?

exonian: Good catch. Will answer this in the PR description.

PR #635, opened 2020-05-28
LGA-1188 review configuration

said-moj:

Removed the following environment variables:
- SET_DEBUG -- was only used in settings/aws.py and settings/docker.py which will be removed
- OBIEE_ENABLED --  only set in settings.py and not used
- IS_AWS -- was being used to determine queue prefix and also which diversity keys to use. Moved to checking if diversity keys env vars are set instead
- APP_NAME -- unused

Statsd configuration was not carried across as there needs to be a separate review if we should continue using this tool

Replica db configuration was not carried across as this feature is not yet available in cloud platforms

Intentionally left blank.

## Checklist


helm_deploy/cla-backend/values-dev.yaml
```diff
@@ -38,24 +26,7 @@ envVars:
...
-  CELERY_PREDEFINED_QUEUE_URL:
-    secret:
-      optional: true
-  OBIEE_ZIP_PASSWORD:
-    secret:
-      optional: true
+
+
+
+
```
exonian, line 32: :D

PR #640, opened 2020-07-03
LGA-1257 - Configure circleci staging/uat deployments

said-moj:

- Created a `multideploy` circleci job that is used in the uat, staging and production workflows.
- Added uat_deploy and production_deploy scripts to deploy to their respective environments
- Added UAT_HOST and PRODUCTION_HOST environment variables (STAGING_HOST already existed)
- Made staging environment use RDS instance instead of local postgres
- Enable ingress on production (was disable and not sure why)

Added a new UAT deployment environment that will be used to deploy feature branches.
Removed approval step from UAT deployment as it's your own environment 
Staging environment will now track master and it's updated before deploying to production
Feature branches will only be deployable to UAT namespace
master branch can be deployed to Staging and Production namespace

## Checklist


.circleci/config.yml
```diff
@@ -138,7 +138,10 @@ jobs:
       - store_artifacts:
           path: test-reports
 
-  staging_deploy:
+  multideploy:
```
exonian, line 141: I would prefer this to just be `deploy`. What's the rationale for `multideploy`? I know that's what we call having multiple UAT environments, but if this step is used for that as well as for the single-deploy environments (staging and prod) then I think it gets confusing.

said-moj: Updated.

I actually was using `deploy` but renamed it to `multideploy` to match what was used in ticket

PR #642, opened 2020-07-10
LGA-1215 - Add replica database config

said-moj:

Add replica database config
Make replica database config optional on UAT and dev

Intentionally left blank.

## Checklist


said-moj:
> So the regular DB has various variables (host, port, password etc).
> Does the replica not need all of these?

No. Replica will have the same username, password and database name as the regular database

However the host will be different. The port can also be different but it's currently setup to use the same 5432 port

exonian:
Ah ok I did wonder if it was that but it just surprised me it would have the same password in particular

PR #643, opened 2020-07-14
LGA-454 Simplify cc=13 cla_backend KnowledgebaseCsvParse#fixture_as_json

mayowaadeniyi:

- Create tests to ensure new functionality completely covers the old functionality and does not break or modify anything
- Reduce cc number by e.g. using in-built pythonic solutions e.g. enumerate over creating a variable to handle index counting
- Break up the functionality and create a function out of it called `_create_article_and_categories`
- `_create_article_and_categories` will create 2 objects: 
1. a record object with a model field name of _knowledgebase.article_ that will be added to the final JSON fixture 
2. an object that will be used by and iterated over later in the fixture_as_json function (line 168) to create record objects with a model field name of _knowledgebase_article_category_matrix_

## Checklist


cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -0,0 +1,1199 @@
...
+                },
+            },
+        ]
+
+        outputList = json.loads(outputJSON)
+        self.compare_length(outputList, expectedList)
+        self.compare_list_of_records(outputList, expectedList)
+
+
+# TODO
```
exonian, line 1194: 👀

exonian:
I think I know what you mean by "native solutions" but it just sounds a bit marketing speak tbh 😅 Probably either refer to them specifically or just as something like "built in language features"

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -0,0 +1,1191 @@
...
+                    u"article_category": 18,
+                    u"preferred_signpost": False,
+                },
+            },
+        ]
+        outputList = json.loads(outputJSON)
+
+        self.compare_length(outputList, expectedList)
+
+        output_records_in_order = outputList[: len(outputList) - 18]
```
exonian, line 1007: You should be able to just use negative indexing here for the slice, ie `[:-18]`

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -0,0 +1,1191 @@
...
+        expected_records_in_order = expectedList[: len(expectedList) - 18]
+
+        output_article_category_maxtrix = outputList[len(outputList) - 18 :]
+        expected_article_category_maxtrix = expectedList[len(expectedList) - 18 :]
+
+        self.compare_list_of_records(output_records_in_order, expected_records_in_order)
+
+        sortedOutputList = sorted(output_article_category_maxtrix, key=lambda x: x["fields"]["article_category"])
+        sortedExpectedList = sorted(expected_article_category_maxtrix, key=lambda x: x["fields"]["article_category"])
```
exonian, line 1016: This is very good, well done.

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -0,0 +1,1191 @@
...
+
+class TestCSV2Fixture(TestCase):
+    def setUp(self):
+        self.datetime_now = datetime.now().replace(tzinfo=timezone.get_current_timezone()).isoformat()
+
+    def compare_dicts(self, outputDict, expectedDict):
+        for key, value in expectedDict.items():
+            self.assertIsInstance(value, type(outputDict[key]))
+            if isinstance(value, int):
+                self.assertEqual(outputDict[key], value)
```
exonian, line 19: This doesn't appear to differ from the fallback comparison (the one behind `else`). Is it needed for making this check go before the datetime one for some reason? Or an oversight?

mayowaadeniyi: Yes it is needed for making it check before datetime. Parse_datetime from the django library only accepts string so when it sees an integer it fails.

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
mayowaadeniyi: Awesome!

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
exonian: Oh I see. Hmm, that's a bit confusing. I think I'd prefer an explicit `elif isinstance(value, string) and isinstance(parse_datetime(value), datetime):` or something. I would never have guess the intent with it as it is

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -0,0 +1,1443 @@
...
+            self.compare_length(outputObject, expectedObject)
+            for key, value in expectedObject.items():
+                self.assertIsInstance(value, type(outputObject[key]))
+                if isinstance(value, dict):
+                    self.compare_dicts(outputObject[key], value)
+                else:
+                    self.assertEqual(outputObject[key], value)
+
+    def compare_length(self, output, expected):
+        self.assertAlmostEqual(len(output), len(expected))
```
exonian, line 56: Why `assertAlmostEqual`?

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -0,0 +1,1443 @@
...
+
+    def compare_article_category_matrices(self, output, expected, pk_range):
+        for outputObj, expectedObj in zip(output, expected):
+            self.compare_length(output, expected)
+            for key, value in expectedObj.items():
+                if key == "pk":
+                    self.assertIsInstance(value, int)
+                    self.assertIsInstance(value, type(outputObj[key]))
```
exonian, line 64: What does this check do? It looks to me like it checks the same thing as the line before but in a more generic way (ie it'd work even if we weren't using ints for pks, but then the line before would fail?)

Actually, I've read the other checks and I think it's just checking that the output and expected values are of the same type? This is confusing for a couple of reasons.
1) The check is that the expected value is the same as the output one. It's not a big thing, but this being back-to-front probably helped confuse me
2) It doesn't actually check the types are the same. It checks the first is the same or some subclass of the second. If you do want to check they're the same, use `assertEqual` instead, with `type()` of both values

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -0,0 +1,1443 @@
...
+                    min_pk_value, max_pk_value = pk_range
+                    self.assertGreaterEqual(value, min_pk_value)
+                    self.assertLessEqual(value, max_pk_value)
+                elif isinstance(value, dict):
+                    self.compare_dicts(outputObj[key], value)
+                else:
+                    self.assertIsInstance(value, type(outputObj[key]))
+                    self.assertEqual(outputObj[key], value)
```
exonian, line 72: All of these lines would be easier to read if we assigned `outputObj[key]` to a variable as soon as we enter this `for` loop (like we did on the previous ticket)

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
mayowaadeniyi: Thanks for that, I forgot I was using assertEqual the whole time while reading the docs..

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
mayowaadeniyi: Done

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
mayowaadeniyi: Actually line 63 doesn't check the values are the same. It just checks the expected value is an integer. That doesn't do anything by itself, so by adding the next check means I can ensure my output from new functionality is an integer.

My issue was we know we want the value to specifically be an integer, hence line 63, but I needed a way to compare that, hence line 64.

Also done it like this because we don't know the actual value.

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
mayowaadeniyi: Done

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
exonian: I didn't suggest to compare the values here

mayowaadeniyi: Ah okay my bad

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
exonian: Is it? I can't find it.

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
exonian: We can talk about this on our next pairing

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
mayowaadeniyi: Ohhh I just noticed the quote is one line, I mis-read. I will change that

exonian: And also says `string` rather than `int`. It's a complete change to the logic of that section

mayowaadeniyi: Now it's done

PR #646, opened 2020-08-12
[WIP] LGA-1349 Ensure JSON fixtures successfully load into DB 

mayowaadeniyi:

- Removed helpline as a field from article records ie. article records no longer have a helpline field
- Removed helpline from CSV files used for testing
- Modified existing tests for helpline field removal
1. Rearranged tests so that the output code was being compared with the expected code rather than the other way around 
- Removed most of the helper test functions as they were redundant
- Create test method for loading JSON fixtures into DB `load_JSON_fixture_into_DB`

_

> Tests were modified to now intentionally test for specific features or values instead of testing for everything. Helper functions were no longer needed as the new tests are more specific and depend on the test case.

_

Steps I took was change expected test results, remove helpline as a field from test CSV files, then finally remove helpline from the actual code functionality.

`A lot of the code changes in the diff are changes to the expected data`

In explaining the **test modifications** further:

1. I did this so that tests would fail until I actually removed the helpline field from the code. Otherwise if I compared expected to output, the expected would not know if helpline was removed as it does not exist in expected and tests would pass
2. I added try - except block statements to the tests because initially upon removing the helpline field from test CSV files, the tests would fail due to a SystemExit Exception with an error code of -1 and exit. The try - except block means upon the tests failing, a more meaningful message can be given.

Another point is I used the `builddata` command in the `load_JSON_fixture_into_DB` method because this generates a JSON fixture then writes it to a file. `loaddata` then takes this file and tries to push it into the DB. 

Other option is to create JSON fixtures from these CSV files, dynamically create .json files from them and then pass these .json files to `loaddata` directly.

Intentionally left blank.

## Checklist


cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -1441,3 +1441,206 @@ def test_fixture_with_empty_csv(self):
         outputList = json.loads(outputJSON)
         self.compare_length(outputList, expectedList)
         self.compare_list_of_records(outputList, expectedList)
+
+    def test_fixture_with_helpline_field_removed(self):
```
exonian, line 1445: Why is this is its own test? Aren't you going to be removing helpline from all your tests?

mayowaadeniyi: Yes looking at it now, what I did doesn't make much sense. Ty

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -12,9 +15,25 @@ class TestCSV2Fixture(TestCase):
     def setUp(self):
         self.datetime_now = datetime.now().replace(tzinfo=timezone.get_current_timezone()).isoformat()
 
-    def calculate_pk_range(self, article_category_matrices):
-        min_range = min(article_category_matrices, key=lambda x: x["pk"])
-        max_range = max(article_category_matrices, key=lambda x: x["pk"])
```
mayowaadeniyi, line 17: I changed the arg name from article_category_matrices to article_category_matrix

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -28,1416 +47,1446 @@ def slice_records(self, output_records, expected_records, start_index=None, end_
...
-        ]
-        outputList = json.loads(outputJSON)
-        self.compare_length(outputList, expectedList)
-        self.compare_list_of_records(outputList, expectedList)
+        csv_file = self.find_path_to_csvfile("legal_resource_entry_type")
+        file = open(csv_file)
+        try:
+            csv = KnowledgebaseCsvParse(file)
+        except SystemExit:
+            self.fail("Missing required fields")
```
exonian, line 476: How do you know this is the case from the exception? Seems like a risky error message

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -28,1416 +47,1446 @@ def slice_records(self, output_records, expected_records, start_index=None, end_
...
+            outputList = json.loads(outputJSON)
 
-        self.compare_length(outputList, expectedList)
+            self.compare_length(outputList, expectedList)
 
-        output_article_categories, expected_article_categories = self.slice_records(
-            outputList, expectedList, end_index=18
-        )
-        self.compare_list_of_records(output_article_categories, expected_article_categories)
+            # Refactor
```
exonian, line 1284: What is this comment doing here?

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
exonian: Why? Are you only passing it one matrix object now? I don't think that's the case...

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -28,1416 +47,1446 @@ def slice_records(self, output_records, expected_records, start_index=None, end_
...
-        self.compare_list_of_records(output_article_1, expected_article_1)
+            output_article_1, expected_article_1 = self.slice_records(outputList, expectedList, 18, 19)
+            self.compare_list_of_records(output_article_1, expected_article_1)
 
-        output_acm_1, expected_acm_1 = self.slice_records(outputList, expectedList, 19, 37)
-        pk_range_acm_1 = self.calculate_pk_range(output_acm_1)
-        output_sorted_acm_1, expected_sorted_acm_1 = self.sort_article_category_matrices(output_acm_1, expected_acm_1)
-        self.compare_article_category_matrices(output_sorted_acm_1, expected_sorted_acm_1, pk_range_acm_1)
+            output_acm_1, expected_acm_1 = self.slice_records(outputList, expectedList, 19, 37)
+            pk_range_acm_1 = self.calculate_pk_range(expected_acm_1)
```
mayowaadeniyi, line 1294: calculate_pk_range only takes in 1 article category matrix

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
mayowaadeniyi: Yes it only takes in 1 article_category_matrix. It was called matrices before because I wanted to use the same name it was called/references to in other files.

This is evidenced by my newest comment below

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
exonian: No, this is still a list of `CategoryMatrix` objects. Min and max pks wouldn't make sense if it was just one object, would it?

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -12,9 +15,25 @@ class TestCSV2Fixture(TestCase):
...
+            management.call_command("builddata", "load_knowledgebase_csv", self.find_path_to_csvfile(relative_path))
+            management.call_command(
+                "loaddata", os.path.abspath("cla_backend/apps/knowledgebase/fixtures/kb_from_spreadsheet.json")
+            )
+        except Exception as e:
+            self.fail(e)
+        else:
+            print("Record successfully loaded into DB!")
+
+    def calculate_pk_range(self, article_category_matrix):
```
exonian, line 34: We discussed how this variable should probably either be plural or have something like `_list` in it

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -28,1416 +47,724 @@ def slice_records(self, output_records, expected_records, start_index=None, end_
...
+                        "when_to_use": u"Baz",
+                        "how_to_use": u"Foo",
+                        "address": u"Foo",
+                        "keywords": u"Bar",
+                        "opening_hours": u"Bar",
+                    },
+                }
+            ]
+            outputList = json.loads(outputJSON)
+            outputArticle = filter(lambda x: x["model"] == "knowledgebase.article", outputList)
```
exonian, line 304: This is nice

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -28,1416 +47,724 @@ def slice_records(self, output_records, expected_records, start_index=None, end_
...
-            },
-        ]
-        outputList = json.loads(outputJSON)
+    def test_fixture_handling_with_entry_type_of_other_resource_for_clients(self):
+        csv_file = self.find_path_to_csvfile("csv_with_entry_type")
+        file = open(csv_file)
+        try:
+            csv = KnowledgebaseCsvParse(file)
+        except SystemExit:
+            self.fail("Missing required fields")
```
exonian, line 328: I still think this assumption is risky

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -28,1416 +47,724 @@ def slice_records(self, output_records, expected_records, start_index=None, end_
...
 
-        output_article_2, expected_article_2 = self.slice_records(outputList, expectedList, 37, 38)
-        self.compare_list_of_records(output_article_2, expected_article_2)
+            article_records = sorted_records["knowledgebase.article"]
+            article_1, article_2 = article_records
 
-        output_article_2_category_matrix = outputList[-18:]
-        expected_article_2_category_matrix = expectedList[-18:]
+            self.assertEqual(len(article_records), 2)
+            self.assertNotEqual(article_1["fields"]["website"], article_2["fields"]["website"])
```
exonian, line 577: I know our intent is to ensure that we aren't creating duplicate output, but I think it'd be better communicated by checking that `article_1["fields"]["website"]` is what we expect it to be _and_ `article_2["fields"]["website"] is what we expect it to be. Checking they are different from each other assumes they're different in the input, and it's not really clear why we expect them to differ from each other.

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -28,1416 +47,724 @@ def slice_records(self, output_records, expected_records, start_index=None, end_
...
+            self.assertNotEqual(article_1["pk"], article_2["pk"])
```
exonian, line 578: Whereas I think this is a good assertion, because checking they're different from each other makes sense with no further information

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -28,1416 +47,724 @@ def slice_records(self, output_records, expected_records, start_index=None, end_
...
+
+            self.assertEqual(len(article_category_matrices), 36)
+            self.assertEqual(min_pk, 1)
+            self.assertEqual(max_pk, 36)
+
+            first_set_of_acm = filter(lambda x: x["fields"]["article"] == 1, article_category_matrices)
+            second_set_of_acm = filter(lambda x: x["fields"]["article"] == 2, article_category_matrices)
+
+            self.assertEqual(len(first_set_of_acm), 18)
+            self.assertEqual(len(second_set_of_acm), 18)
```
exonian, line 591: I like these too, they read very cleanly 💪

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -28,1416 +47,724 @@ def slice_records(self, output_records, expected_records, start_index=None, end_
...
-                },
-            },
-        ]
+        csv_file = self.find_path_to_csvfile("empty_csv")
+        file = open(csv_file)
+        try:
+            csv = KnowledgebaseCsvParse(file)
+        except SystemExit:
+            self.fail("SystemExit Exception with a error code of -1")
+        else:
```
exonian, line 600: I would probably keep this test but just assert how many objects are described (rather than checking their actual data), so just check there are 18 categories and no articles or matrices

cla_backend/apps/knowledgebase/tests/test_csv_2_fixture.py
```diff
@@ -28,1416 +47,724 @@ def slice_records(self, output_records, expected_records, start_index=None, end_
...
+        csv_file = self.find_path_to_csvfile("complete_article")
+        file = open(csv_file)
+        try:
+            csv = KnowledgebaseCsvParse(file)
+        except SystemExit:
+            self.fail("Missing required fields")
+        else:
+            outputJSON = csv.fixture_as_json()
+            outputList = json.loads(outputJSON)
+            expectedList = [
```
exonian, line 345: I would probably prefer a check of just the attributes you care about, for previously-discussed reasons of communicating intent. I wonder if you could restructure this as either looping through the output and asserting the values you want to assert or -- and this second one is probably preferable -- converting expectedList into a list of just the values you want to check (probably `article`, `article_category` and `preferred_signpost`) and then you can assert that that list matches a defined output (which would then be much easier to read than our current expected output)

exonian (changes requested):
Getting there, looking really, really good

exonian (approved):
This is absolutely fantastic work.

You've taken on my advice, you've made some really neat changes like the path handling and the `assertItemsEqual()`, you've tidied up your helper functions now they're not needed, and you've really thought about what you're testing and why.

PR #651, opened 2020-09-02
LGA-1350 - Allocation tests

exonian:
Adds tests to demonstrate the allocation process across days (where currently only cases from the same day are taken into account). Skips the test that would require this to be changed.

Follows @said-moj's fix of other tests in order to get a previous test working, and reduces its volume from 8000 cases to 80 for practicality and because there's no reason to have so many.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ pip-log.txt
 .coverage
 .tox
 nosetests.xml
+test-reports
 
 # Translations
 *.mo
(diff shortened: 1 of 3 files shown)


said-moj:
@exonian through these tests did we discover reasons for the distribution anomalies

exonian:
@said-moj Yeah, it illustrated to me that the allocations would deviate from equal if the volumes per day don't divide equally, and that these deviations are considerable at low volumes.

said-moj (approved):

said-moj:
> @said-moj Yeah, it illustrated to me that the allocations would deviate from equal if the volumes per day don't divide equally, and that these deviations are considerable at low volumes.

Non-blocking: Would be good to have this on the ticket

PR #653, opened 2020-09-11
LGA-1383 - Remove call back scheduling delay

exonian:
Allows callbacks to be scheduled in the current and next callback windows, in addition to all future windows as normal.

Requires corresponding frontend change in order to provide its benefit. cla_public already enforces stricter rules than this (it doesn't allow callbacks within 2 hours), so its behaviour won't be affected.

## Checklist

### DIFF
--- a/cla_backend/apps/call_centre/forms.py
+++ b/cla_backend/apps/call_centre/forms.py
@@ -201,7 +201,7 @@ class CallMeBackForm(BaseCallMeBackForm):
     priority_callback = forms.BooleanField(required=False)
 
     def _is_dt_too_soon(self, dt):
-        return dt <= timezone.now() + datetime.timedelta(minutes=30)
+        return dt <= timezone.now() - datetime.timedelta(minutes=30)
 
     def _is_dt_out_of_hours(self, dt):
         _dt = dt
--- a/cla_backend/apps/call_centre/tests/test_forms.py
+++ b/cla_backend/apps/call_centre/tests/test_forms.py
@@ -453,12 +453,10 @@ class CallMeBackFormTestCase(BaseCaseLogFormTestCaseMixin, TestCase):
         # invalid format
         _test(case, now_utc.strftime("%Y/%m/%d %H:%M"), u"Enter a valid date/time.")
 
-        # datetime in the past
-        _test(case, self._strftime(now_utc), u"Specify a date not in the current half hour.")
-
+        # datetime in the past, beyond the current window
         _test(
             case,
-            self._strftime(now_utc + datetime.timedelta(minutes=30)),
+            self._strftime(now_utc - datetime.timedelta(minutes=35)),
             u"Specify a date not in the current half hour.",
         )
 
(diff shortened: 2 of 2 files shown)


cla_backend/apps/call_centre/tests/test_forms.py
```diff
@@ -482,6 +480,28 @@ def _test(case, datetime, error_msg):
...
+
+            self.assertTrue(form.is_valid())
+
+        # datetime in the future
+        _test(case, self._strftime(now_utc + datetime.timedelta(minutes=10)))
+        _test(case, self._strftime(now_utc + datetime.timedelta(minutes=30)))
+        _test(case, self._strftime(now_utc + datetime.timedelta(minutes=180)))
+
+        # datetime in the current 30 minute window
+        _test(case, self._strftime(now_utc - datetime.timedelta(minutes=14)))
```
mayowaadeniyi, line 503: Is line 498 not testing the same thing as line 503?

exonian: Line 503 is in the past, subtracted timedelta. I did wonder if that was easy to miss, and it is, so I'll probably change it to being `+ datetime.timedelta(minutes=-14)` , which I think does the same thing but is easier to spot

mayowaadeniyi: Oh no sorry Michael, I meant aren't both tests testing for the datetime within the same callback window? For example, if the time was 10.15am, wouldn't -14minutes and +10minutes from 10.15am still be within the same callback window?

exonian: Ah yes I guess so, good catch! Do you think I should delete the 10 minute one?

mayowaadeniyi: Yeah I think with the comments it would make sense to delete the 10min one. To nitpick, could we change -14mins to -10mins just so it's more obvious?

cla_backend/apps/call_centre/forms.py
```diff
@@ -201,7 +201,7 @@ class CallMeBackForm(BaseCallMeBackForm):
     priority_callback = forms.BooleanField(required=False)
 
     def _is_dt_too_soon(self, dt):
-        return dt <= timezone.now() + datetime.timedelta(minutes=30)
+        return dt <= timezone.now() - datetime.timedelta(minutes=30)
```
mayowaadeniyi, line 204: Isn't this change the opposite of what we want as it will not allow a datetime to be set unless it has been over 30mins from the current datetime?

cla_backend/apps/call_centre/tests/test_forms.py
```diff
@@ -453,12 +453,10 @@ def _test(case, datetime, error_msg):
         # invalid format
         _test(case, now_utc.strftime("%Y/%m/%d %H:%M"), u"Enter a valid date/time.")
 
-        # datetime in the past
-        _test(case, self._strftime(now_utc), u"Specify a date not in the current half hour.")
-
```
mayowaadeniyi, line 458: Isn't this test fine? Can we not change the comment to `datetime in the current callback window` or `datetime which is too early`

exonian: No, that would fail. That "comment" is the expected error message when running `_test()`. With the new rules, we would not expect an error when using the current datetime.

cla_backend/apps/call_centre/forms.py
exonian: No to the first question, of what this does: it doesn't change the direction of the check (it's still "must be after) but the time it has to be after is now in the past not the future.

Yes to the second question, that is exactly what it would do. At 10:31, 10:00-10:30 is not valid but 10:30-11:00 is valid

cla_backend/apps/call_centre/tests/test_forms.py
exonian: I don't see what that would achieve

cla_backend/apps/call_centre/forms.py
mayowaadeniyi: Ah okay, I think I understand now. Just to confirm, the first callback slot available at 10.31 would be 10.30 - 11am? On Slack, you said 11 - 11.30am

cla_backend/apps/call_centre/tests/test_forms.py
mayowaadeniyi: Yeah I realised it doesn't matter. What the test is doing is more important + the comment is already there

cla_backend/apps/call_centre/forms.py
exonian: I don't think I did.

cla_backend/apps/call_centre/tests/test_forms.py
exonian: Actually, I'm going to leave in both the past-but-same-window and future-but-same-window, as I think having them both is slightly valuable against regression, but it was a very good catch 👍

said-moj (approved):
👍 LGTM

mayowaadeniyi (approved):
LGTM 👍  also, plus I need to approve for this to go through anyway..

PR #654, opened 2020-09-18
LGA-1396 - Fix allocations across days

exonian:
Removes the constraint that only considers the current day when deciding whether to balance up the allocations or allocate randomly.
There is already a constraint in there to limit it to allocations since the weightings were changed.

Intentionally left blank.

## Checklist


said-moj (approved):
👍 LGTM

PR #659, opened 2020-09-24
LGA-1433 - Fix admin images

exonian:
Fixes static files, which come from s3, by setting AWS_QUERYSTRING_AUTH to False.
Static files bucket is public so doesn't need signed urls. As they'd
expire, and because it makes paths harder to naively construct, we're
much better off without (right now, the signature breaks our image paths
and turning it off fixes them).

The reports do not use signed urls for access, so are not affected by
this

## Checklist


said-moj (approved):
👍 LGTM

PR #660, opened 2020-09-24
LGA-1434 - Stop non-UAT releases using unique names

exonian:
Sets $RELEASE_NAME to 'cla-backend' on staging and production, so that
the Helm chart is consistently named and therefore future releases
update the chart regardless of branch name (although the branch name
will usually be consistent for these deploy types anyway).

Removes the fullnameOverride on staging and production, as we don't want
or need our Kubernetes objects to have branch-specific names - again,
that's a UAT environment thing allowing multiple releases to exist at
once, and is detrimental on staging and production for similar reasons
to the release naming. Without the override, these releases will default
to being called 'cla-backend', from the Chart.yaml definition, and this
is what will be the basis of the naming of the deployments, ingresses
etc created for the release.

Intentionally left blank.

## Checklist

### DIFF
--- a/.circleci/define_build_environment_variables
+++ b/.circleci/define_build_environment_variables
@@ -25,9 +25,10 @@ case $NAMESPACE in
 esac
 
 export CLEANED_BRANCH_NAME=$(echo $CIRCLE_BRANCH | sed 's/^feature[-/]//' | sed 's:^\w*\/::' | tr -s ' _/[]().' '-' | tr '[:upper:]' '[:lower:]' | cut -c1-28 | sed 's/-$//')
-export RELEASE_NAME=${CLEANED_BRANCH_NAME}
 if [ $NAMESPACE == "uat" ]; then
+  export RELEASE_NAME=${CLEANED_BRANCH_NAME}
   export RELEASE_HOST=${RELEASE_NAME}-${NAMESPACE_HOST}
 else
+  export RELEASE_NAME=cla-backend
   export RELEASE_HOST=${NAMESPACE_HOST}
 fi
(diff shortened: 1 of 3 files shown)


said-moj:
@exonian when this is merged and deployed to production what happens to the current deployment?
Which deployment would the ingress refer to?

exonian:
@said-moj Good point. There would be a second helm release, with all its accompanying stuff, including a second ingress. So we may want to do this slightly out-of-hours, just to be doubly sure, as the old chart will need to be uninstalled.

On staging, we currently have two ingresses, which I can only assume are fighting in some non-ideal way...

Because the release name is being used to prefix all the objects (we wrote that in order to make multideploys work), the ingress-service-deployment-pod chains should all be fine and not tangle, but we would have both ingresses because they have different names: it doesn't care if they have the same hostname, apparently.

exonian:
Or, we could rename the existing Ingress to `cla-backend`, leaving it connected to the branch-named Service. Then when we deployed, it would replace the Ingress. That could be the best play.

said-moj:
> Or, we could rename the existing Ingress to `cla-backend`, leaving it connected to the branch-named Service. Then when we deployed, it would replace the Ingress. That could be the best play.

Can we try this on staging, if it works on staging we might still want to do it out of hours to be doubly sure

exonian:
> Can we try this on staging, if it works on staging we might still want to do it out of hours to be doubly sure

Yep sounds like a plan. Shall we override the circleci branch rules to allow us to deploy to staging before merging this to the epic branch?

said-moj:
> > Can we try this on staging, if it works on staging we might still want to do it out of hours to be doubly sure
> 
> Yep sounds like a plan. Shall we override the circleci branch rules to allow us to deploy to staging before merging this to the epic branch?

Yes please

exonian:
Ah it turns out you can't edit the `name` attribute, so we'll need to do a deploy and then use the chart uninstall to remove the previous Ingress. Which will still be fine.

said-moj (approved):

PR #662, opened 2020-10-06
LGA-1449 - Add readiness/liveness probes to worker pods

said-moj:

Add readiness/liveness probes to worker pods

Usually the status of celery can be checked by running `celery -A cla_backend status` however there is a general issue with running any of the celery inspect commands, which is celery attempts to create `<uuid>.celery.pidbox` queues that it uses for broadcasting messages however due to the restrictive permissions on the aws sqs (cannot create new queues) this fails.

I couldn't disable these broadcasts or set the name of the queue that it broadcasts to.

## Checklist

### DIFF
--- a/cla_backend/apps/core/management/commands/worker_probes.py
+++ b/cla_backend/apps/core/management/commands/worker_probes.py
@@ -1,0 +1,20 @@
+from django.conf import settings
+from django.core.management.base import NoArgsCommand
+from cla_backend.celery import app as celery
+from legalaid.models import Case
+
+
+class Command(NoArgsCommand):
+    def handle_noargs(self, **options):
+        self.celery_probe()
+        self.database_probe()
+
+    def database_probe(self):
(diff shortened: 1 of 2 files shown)


exonian (approved):
Timings might need tweaking: see how it goes on staging. But looks good to me

exonian:
Oh it's probably the hyphen in the command name. Might be better off with an underscore

cla_backend/apps/core/management/commands/worker_probes.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Command(NoArgsCommand):
+    def handle_noargs(self, **options):
+        self.celery_probe()
+        self.database_probe()
+
+    def database_probe(self):
+        # Ability to connect to the db; doesn't matter if the result is True or False
+        #
```
exonian, line 14:
```suggestion
```

exonian (approved):

PR #663, opened 2020-10-06
LGA-1450 - Sentry SDK

exonian:
Switches from Raven library to Sentry SDK. Uses a kubernetes secret that is already present, making switchover from the old Sentry instance to the new one easy.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/celery.py
+++ b/cla_backend/celery.py
@@ -4,10 +4,7 @@ import os
 
 from celery import Celery
 from django.conf import settings
-from raven.contrib.celery import register_logger_signal, register_signal
 
-# set the default Django settings module for the 'celery' program.
-from raven.scripts.runner import send_test_message
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "cla_backend.settings")
 app = Celery("cla_backend")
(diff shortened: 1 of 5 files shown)


said-moj:
Is there any reason why we have to remove the middlware([1](https://github.com/ministryofjustice/cla_backend/pull/663/files#diff-6e46e10f71a9a0c60a5b076728bbb889L275),[2](https://github.com/ministryofjustice/cla_backend/pull/663/files#diff-569c891944183a97e7caa899c62ca3e3L28)) and [error handler](https://github.com/ministryofjustice/cla_backend/pull/663/files#diff-6e46e10f71a9a0c60a5b076728bbb889L260)

exonian:
@said-moj I must admit, I had missed that the sentry middleware we were using wasn't just the standard one. But are we even doing anything with that message ID stuff?

As for the `Sentry()` bit being removed, I just don't think that's how the new sdk does it, I think the `sentry_sdk.init()` in settings should be enough.

Likewise I don't think we need that log handler defined; I'm not even convinced it was doing anything, as no `logger` uses it, but maybe I'm missing something on that?

said-moj:
>But are we even doing anything with that message ID stuff?
Honestly, I am not sure. I am none the wiser with these sentry stuff

said-moj:
Do you think we could add 404 logging https://docs.sentry.io/clients/python/integrations/#404-logging

I guess it would be good to know what's 404'ing given this project is mostly api based, or maybe we could add that as another ticket?

said-moj:
I have created a separate ticket for the 404 logging https://dsdmoj.atlassian.net/browse/LGA-1466

said-moj (approved):
👍 LGTM

PR #665, opened 2020-10-08
LGA-1460 - uwsgi graceful termination workaround

said-moj:
Add preStop lifecycle hook and terminationGracePeriod to allow enough time for uwsgi finish processing requests

Changes explained:

- **die-on-term** - respect the convention of shutting down the instance
- **preStop** - Sleep for 30 seconds. Called before receiving SIGTERM. Aim here is to sleep enough time for the current requests to finish.
- **terminationGracePeriodSeconds** - Amount of time kubernetes waits before sending us the SIGKILL signal.

Important note from the [kubernetes documentation ](https://cloud.google.com/blog/products/gcp/kubernetes-best-practices-terminating-with-grace)on how the preStop and terminationGracePeriodSeconds interact:

> It’s important to note that this(_terminationGracePeriodSeconds_) happens in parallel to the preStop hook and the SIGTERM signal. Kubernetes does not wait for the preStop hook to finish.
If your app finishes shutting down and exits before the terminationGracePeriod is done, Kubernetes moves to the next step immediately.

uwsgi seems to be brutally reloading when receiving the SIGTERM signal which kills any current requests.
This workaround attempts to give it enough time to finish processing the current requests

## Checklist

### DIFF
--- a/helm_deploy/cla-backend/templates/deployment.yaml
+++ b/helm_deploy/cla-backend/templates/deployment.yaml
@@ -22,6 +22,7 @@ spec:
       serviceAccountName: {{ include "cla-backend.serviceAccountName" . }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      terminationGracePeriodSeconds: 30
       containers:
         - name: {{ .Chart.Name }}
           securityContext:
@@ -46,6 +47,10 @@ spec:
               httpHeaders:
                 - name: Host
                   value: "{{ .Values.host }}"
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sleep","30"]
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           env:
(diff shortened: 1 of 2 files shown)


helm_deploy/cla-backend/templates/deployment.yaml
```diff
@@ -22,6 +22,7 @@ spec:
       serviceAccountName: {{ include "cla-backend.serviceAccountName" . }}
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      terminationGracePeriod: 30
```
exonian, line 25: Sorry, my bad, I think this should be `terminationGracePeriodSeconds`

said-moj:
I tested by having the following script as the preStop script, test results can be seen in project sentry:
`import os
import time
import datetime
from raven import Client

start = datetime.datetime.now()
time.sleep(30)
end = datetime.datetime.now()
client = Client(os.environ.get("RAVEN_CONFIG_DSN"))
message = "POD-PRESTOP\nStart: {}\nFinish:{}\n".format(start, end)
client.captureMessage(message)
`

exonian (approved):

PR #671, opened 2020-10-14
LGA-1476 - Add cronjob for send_to_performance_platform

exonian:
Adds a cronjob to the cronjobs kubernetes definition, to run the send_to_performance_platform management command, weekly on Mondays at 01:01, as we did previously via the salt stack on cla_backend-deploy.
The environment variables needed by the command are already in place.

Also rename the containers of all the cronjob pods to be consistent,
rather than each having the name of the job

## Checklist


said-moj (approved):
👍  LGTM

PR #672, opened 2020-10-19
LGA-1481 - Persistent UAT env

exonian:
Adds a "static UAT deploy" step to the Circle pipeline, with its own approval step.
Modifies the `define_build_environment_variables` script to take a second argument and use it as `DYNAMIC_HOSTNAME` to decouple the multideploy UAT behaviour from the UAT namespace (as the new release needs to be in the latter but not use the former), and modifies the pipeline to have this as a build job parameter and pass it through.

Intentionally left blank.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,6 +142,8 @@ jobs:
     parameters:
       namespace:
         type: string
+      dynamic_hostname:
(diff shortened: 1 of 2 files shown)


said-moj (approved):
LGTM 👍

PR #675, opened 2020-11-04
Static UAT use cloud platform database

said-moj:

- Move local postgress credentials from secret to values file. The database secret will be populated with the persistent cloud platform hosted postgres credentials

Intentionally left blank.

## Checklist

### DIFF
--- a/helm_deploy/cla-backend/templates/deployment.yaml
+++ b/helm_deploy/cla-backend/templates/deployment.yaml
@@ -104,11 +104,7 @@ spec:
             {{- toYaml .Values.resources | nindent 12 }}
           env:
             - name: POSTGRES_PASSWORD
-              valueFrom:
-                secretKeyRef:
-                  name: database
-                  key: password
-                  optional: true
+              value: {{ .Values.envVars.DB_PASSWORD.value }}
             - name: POSTGRES_USER
               value: postgres
             - name: POSTGRES_DB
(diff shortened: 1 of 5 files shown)


helm_deploy/cla-backend/templates/deployment.yaml
```diff
@@ -104,11 +104,7 @@ spec:
             {{- toYaml .Values.resources | nindent 12 }}
           env:
             - name: POSTGRES_PASSWORD
-              valueFrom:
```
exonian, line 107: I don't understand how removing this doesn't break it for most of our environments. What am I missing?

said-moj: Yes this has broken existing UAT deployments, they would need to rebase from base branch once this goes in.

However it should not break any deployments on staging/production because this only affects deployments that have a local postgres enabled (It's wrapped in `{{- if .Values.localPostgres.enabled -}}`)

said-moj: Also `valueFrom` is only used when referencing secrets.

You can use `Value` when referencing a static(or templated) value.

exonian: > this only affects deployments that have a local postgres enabled (It's wrapped in {{- if .Values.localPostgres.enabled -}})
Thanks, this was the bit I was missing 👍

exonian (approved):

PR #676, opened 2020-11-05
LGA-1466 - Sentry log 404s

exonian:

Logs to sentry. It only creates one event, regardless of path, but the path is broken up and available as event tags so we should be able to manage the different 404s on the Sentry platform by custom grouping and ignoring rules.

Intentionally left blank.

## Checklist


said-moj:
Is merging into master intentional?

said-moj (approved):
👍 LGTM

PR #679, opened 2020-11-30
LGA-1461 - Remove checking if we are on the first EC2 instance

said-moj:

Remove checking if we are on the first ec2 instance

Application has moved to kubernetes and these management commands are no longer run by uwsgi 
Instead they are run by a dedicated kubernetes job

## Checklist


mayowaadeniyi (approved):
LGTM 👍

PR #681, opened 2020-12-04
LGA-1542 - Remove mortgage cap

said-moj:

- Remove mortgage cap
- Introduced MORTGAGE_CAP_REMOVAL_DATE setting with the date the mortgage cap is to be removed.
- **MORTGAGE CAP WILL AUTOMATICALLY BE REMOVED AFTER THE DATE IN settings.MORTGAGE_CAP_REMOVAL_DATE**
- Updated affected unittests with dictionary of pre/post mortgage cap removal expected values.

- Remove mortgage cap for all properties(main and additional properties)
- Mortgage cap removal is enabled on static-uat environment

## Checklist

### DIFF
--- a/cla_backend/libs/eligibility_calculator/tests/test_calculator.py
+++ b/cla_backend/libs/eligibility_calculator/tests/test_calculator.py
@@ -13,7 +13,6 @@ from ..models import CaseData, Facts
 class MortgageCapRemovalMixin(object):
     def setUp(self):
         super(MortgageCapRemovalMixin, self).setUp()
-        super(MortgageCapRemovalMixin, self).setUp()
         if CapitalCalculator.is_post_mortgage_cap_removal():
             self.expected_results_key = "post_mortgage_cap_removal"
         else:
--- a/cla_backend/libs/eligibility_calculator/tests/test_calculator.py
+++ b/cla_backend/libs/eligibility_calculator/tests/test_calculator.py
@@ -63,10 +63,10 @@ class TestCapitalCalculator(unittest.TestCase):
         calc = CapitalCalculator(properties=[self.make_property(52000000, 15000000, 100, True, True)])
         capital = calc.calculate_capital()
 
-        self.assertEqual(capital, 22000000)
-        self.assertEqual(calc.main_property["equity"], 22000000)
+        self.assertEqual(capital, 17000000)
+        self.assertEqual(calc.main_property["equity"], 17000000)
         self.assertDictEqual(
-            calc.calcs, {"property_equities": [22000000], "property_capital": 22000000, "liquid_capital": 0}
+            calc.calcs, {"property_equities": [17000000], "property_capital": 17000000, "liquid_capital": 0}
         )
 
     def test_scenario_smod_3(self):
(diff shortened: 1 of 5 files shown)


cla_backend/libs/eligibility_calculator/tests/test_calculator.py
```diff
@@ -10,7 +10,28 @@
 from ..models import CaseData, Facts
 
 
-class TestCapitalCalculator(unittest.TestCase):
+class MortgageCapRemovalMixin(object):
+    def setUp(self):
+        super(MortgageCapRemovalMixin, self).setUp()
+        super(MortgageCapRemovalMixin, self).setUp()
```
exonian, line 16: Please tell me this duplication is just a mistake and not needed for some horrible reason... 😅

cla_backend/libs/eligibility_calculator/tests/test_calculator.py
```diff
@@ -773,47 +924,65 @@ def test_pensioner_300k1p_house_100k1p_mort_799999_savings(self):
...
                 "disposable_capital_assets": 80001,
             },
-        )
+            "post_mortgage_cap_removal": {
+                "property_capital": 10000001,
+                "property_equities": [10000001],
+                "disposable_capital_assets": 80000,
+            },
+        }
+        expected_results.update(expected_property_results[self.expected_results_key])
```
exonian, line 982: This is really nicely done

cla_backend/settings/base.py
```diff
@@ -369,6 +369,9 @@
 CONTRACT_2018_ENABLED = os.environ.get("CONTRACT_2018_ENABLED", "True") == "True"
 PING_JSON_KEYS["CONTRACT_2018_ENABLED_key"] = "CONTRACT_2018_ENABLED"
 
+MORTGAGE_CAP_REMOVAL_DATE = os.environ.get("MORTGAGE_CAP_REMOVAL_DATE", "2021-01-28 00:00")
```
exonian, line 372: This works, and I can see the rationale for it (first var matches env var, and we want the env var name to match the final setting name too) but I really don't like it. Maybe hypocritical as I've probably written things like this, but I think I'd much prefer the initial one to have a different name, and probably lowercased in normal Python variable style.
So like
```
mortgage_cap_removal_date = os.environ.get("MORTGAGE_CAP_REMOVAL_DATE", "2021-01-28 00:00")
MORTGAGE_CAP_REMOVAL_DATE = datetime.datetime.strptime(mortgage_cap_removal_date, "%Y-%m-%d %H:%M")
```
or, alternatively, a name that differs by more than just case (maybe with "_string" in?)

exonian (commented):
This is looking really good. Couple of queries / soft requests for change

cla_backend/libs/eligibility_calculator/tests/test_calculator.py
said-moj: Good catch. Yep it's a mistake

cla_backend/settings/base.py
said-moj: I agree, you're example seems much more readable

exonian (approved):

PR #683, opened 2020-12-18
LGA-1472 - Knowledgebase articles export

exonian:
Adds a new report to output the knowledgebase articles information. This report uses the ORM, which I think makes the relationships (phone numbers and categories) stuff easier, or at least it makes it easier to understand with the same core knowledge as the bulk of the repo.

I've written tests for the new report. These illustrate that it requires a few SQL queries in order to get everything ready, but this doesn't increase with the number of records or anything like that, and each query should be quick.

## Checklist


said-moj (approved):
LGTM 👍

PR #687, opened 2021-01-08
LGA-1560 - Fix Celery on UAT multideploy environments

exonian:
- Makes the s3 setup a bit easier to use locally by rearranging some setup
- Makes the sqs config clearer by changing the way the settings are combined
- Makes multideploy UAT environments use Celery in synchronous mode to work around there only being one SQS available to the namespace, and stops them creating worker pods

This was all written originally as part of #683 but that's not ready to go yet so it's getting its own PR now. 

## Checklist


said-moj (approved):
Great work 👍

PR #689, opened 2021-01-14
LGA-815: Slow complaints tab

Pheonnexx:

This amends the SQL query that runs when in the frontend dashboard they are viewing all complaints that are not closed or voided.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/complaints/views.py
+++ b/cla_backend/apps/complaints/views.py
@@ -95,19 +95,8 @@ class BaseComplaintViewSet(
             }
         )
         if dashboard and not show_closed:
-            qs = qs.extra(
-                where=[
-                    """
-                    SELECT COUNT(id) < 1 FROM cla_eventlog_log
-                    WHERE
-                        cla_eventlog_log.content_type_id={complaint_ct}
-                        AND cla_eventlog_log.object_id=complaints_complaint.id
-                        AND cla_eventlog_log.code IN ('{closed_code}', '{voided_code}')
-                    """.format(
-                        **sql_params
-                    )
-                ]
-            )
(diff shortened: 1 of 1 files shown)


cla_backend/apps/complaints/views.py
```diff
@@ -95,19 +95,8 @@ def get_queryset(self, dashboard=False, show_closed=False):
...
-                    WHERE
-                        cla_eventlog_log.content_type_id={complaint_ct}
-                        AND cla_eventlog_log.object_id=complaints_complaint.id
-                        AND cla_eventlog_log.code IN ('{closed_code}', '{voided_code}')
-                    """.format(
-                        **sql_params
-                    )
-                ]
-            )
+            complaint_events = ComplaintLog.objects.filter(content_type=sql_params['complaint_ct'], code__in=["COMPLAINT_CLOSED", "COMPLAINT_VOID"]).values_list('object_id', flat=True)
```
said-moj, line 98: This is much simpler to read and comprehend.

Was the original query doing a count and checking if that count was less than 1, and then producing a True or False value?

Pheonnexx: For each complaint I think it was going off and doing a count.  First attempt I changed to using exists which did reduce the time, but not significantly.  Next step was to split apart as I have done here, although definitely learnt about query sets here as there were a couple of ways that the above could have been written that produced slightly different SQL.  

Happy to discuss my findings!

exonian: Yeah I wasn't really surprised that the count->exists change did very little, as I suspected it was mostly that it was doing _something_ for every single one. We can make this bit `ComplaintLog.objects.filter(content_type=sql_params['complaint_ct'], code__in=["COMPLAINT_CLOSED", "COMPLAINT_VOID"])` more reusable next time (if we do similar changes elsewhere), but don't need to for now as it would just be code organisation and not actually change the SQL at all.

said-moj (approved):
👍 LGTM

PR #691, opened 2021-01-22
LGA-1121-Old complaints appearing

Pheonnexx:

The was an issue with old complaints appearing in the complaints list.  What was found was a task called DeleteOldData which was a task that would regularly delete data over two years old, with a recent introduction of audit logs this was no longer working and half deleting data which meant that complaints were now being displayed.

This pull request deletes the audit logs for cases and complaints so that it can successfully delete all required data.

What this should also do on the first run is delete all data that hasn't been deleted correctly.

I have put a distinct filter on the complaints list when no closed complaints are closed.

## Checklist


said-moj:
I like the freezegun package, good find

said-moj (approved):
👍 LGTM

PR #692, opened 2021-02-01
LGA-1599 - Scale prod up to 5 pods

exonian:
Hopefully will help with CHS timeouts and gateway errors.
This has already been applied in production. It is too early to tell if it's really helped, but it needs committing to keep it in place so that we can monitor it.

Intentionally left blank.

## Checklist


Pheonnexx (approved):
LGTM

PR #693, opened 2021-02-02
LGA-1558 - Knowledgebase articles bulk CSV import

said-moj:

Add knowledgebase.models.Article import from CSV

There is already existing functionality to export knowledge base articles as a CSV. This PR introduces a way to re-import that CSV after it has been updated. So the process becomes:

- The user downloads the knowledge base articles export (/admin/reports/all-knowledgebase-articles/)
- Make their changes to the CSV
- Upload the CSV back (/admin/knowledgebase/article/import-csv/)

The first column of every row is the ID(pk) of the Article to make the changes to.
If the row to be imported is a new Article then the ID column must have a value of `NEW`

The KnowledgebaseCSVImporter.save uses a database transaction to make sure all the articles are written to database and if there is an error it should not commit the transaction.

## Checklist

### DIFF
--- a/cla_backend/apps/knowledgebase/tests/test_csv_user_import.py
+++ b/cla_backend/apps/knowledgebase/tests/test_csv_user_import.py
@@ -1,0 +1,137 @@
+import datetime
+from django.test import TestCase
+from knowledgebase.utils.csv_user_import import KnowledgebaseCSVImporter
+from knowledgebase.utils.csv_user_import_mappings import (
+    ARTICLE_COLUMN_FIELD_MAPPING,
+    TELEPHONE_COLUMN_FIELD_MAPPING,
+    ARTICLE_CATEGORY_MATRIX_COLUMN_FIELD_MAPPING,
+)
+from core.tests.mommy_utils import make_recipe
+
+
+class KnowledgebaseCSVImporterTester(TestCase):
+    def setUp(self):
+        self.importer = KnowledgebaseCSVImporter(None)
+        self.row = self.get_csv_row()
+
+    def test_article_row_import__new_article(self):
+        article = self.importer.get_article_from_row(ARTICLE_COLUMN_FIELD_MAPPING, self.row)
+        # Set the pk field in the csv row to None
+        self.row[0] = None
+        self._assert_row_against_article(self.row, article)
+
+    def test_article_row_import__existing_article(self):
+        existing_article = make_recipe("knowledgebase.article")
+        # Set the pk field in the csv row to to an existing Article pk
+        self.row[0] = existing_article.pk
+        article = self.importer.get_article_from_row(ARTICLE_COLUMN_FIELD_MAPPING, self.row)
+        self._assert_row_against_article(self.row, article)
+
+    def test_telephone_numbers_row(self):
+        article = make_recipe("knowledgebase.article")
+        telephone_numbers = self.importer.get_telephone_numbers_from_row(
+            article, TELEPHONE_COLUMN_FIELD_MAPPING, self.row
+        )
+        self.assertEqual(len(telephone_numbers), 2)
+        self._assert_row_telephone(self.row, TELEPHONE_COLUMN_FIELD_MAPPING[0], telephone_numbers[0])
+        self._assert_row_telephone(self.row, TELEPHONE_COLUMN_FIELD_MAPPING[1], telephone_numbers[1])
+
+    def test_article_category_matrices_from_row(self):
+        employment_category = make_recipe("knowledgebase.article_category", name="Employment")
+        discrimination_category = make_recipe("knowledgebase.article_category", name="Discrimination")
+        article = make_recipe("knowledgebase.article")
+        matrices = self.importer.get_get_article_category_matrices_from_row(
+            article, ARTICLE_CATEGORY_MATRIX_COLUMN_FIELD_MAPPING, self.row
+        )
+        self.assertEqual(len(matrices), 2)
+        self.assertEqual(matrices[0].article_category, employment_category)
+        self.assertFalse(matrices[0].preferred_signpost)
+        self.assertEqual(matrices[1].article_category, discrimination_category)
+        self.assertTrue(matrices[1].preferred_signpost)
+
+    def test_article_category_matrices_from_row__non_existing_category(self):
+        article = make_recipe("knowledgebase.article")
+        with self.assertRaisesMessage(ValueError, "Could not find category with name: Employment"):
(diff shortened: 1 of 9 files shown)


cla_backend/apps/knowledgebase/tests/test_csv_user_import.py
```diff
@@ -0,0 +1,215 @@
...
+class KnowledgebaseCSVImporterTester(TestCase):
+    def setUp(self):
+        self.importer = KnowledgebaseCSVImporter()
+        self.row = self.get_csv_row()
+
+    def test_article_row_import__new_article(self):
+        article = self.importer.get_article_from_row(ARTICLE_COLUMN_FIELD_MAPPING, self.row)
+        # We expect the pk value of the new article to be None
+        self.row[0] = None
+        self._assert_row_against_article(self.row, article)
```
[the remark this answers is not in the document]
exonian: Looking at the other cases, you don't need to do that for those, I don't think. Only when the order is use, modify, compare. Most of them are modify, use, compare, and that makes sense with them staying as the same variable

cla_backend/apps/knowledgebase/tests/test_csv_user_import.py
```diff
@@ -0,0 +1,215 @@
...
+            "2015-01-28 15:06:18.718000+00:00",
+            "LEGAL",
+            "LAA",
+            "[PREF DISCRIMINATION]",
+            "LAA (Legal Aid Agency)",
+            "https,//www.gov.uk/government/organisations/legal-aid-agency",
+            "",
+            "This is the article description",
+            "This is the article description",
+            "This ios how to use it",
```
exonian, line 182:
```suggestion
            "This is how to use it",
```

exonian (commented):
Will finish reading the tests in a bit, thought I'd send my feedback so far

cla_backend/apps/knowledgebase/utils/csv_user_import.py
```diff
@@ -0,0 +1,138 @@
...
+                rows.append(self.process_row(row))
+            except ValidationError as e:
+                for field, message in e.message_dict.items():
+                    errors.append("row %s: %s: %s" % (index + 1, field, message))
+            except Exception as e:
+                errors.append("row %s: %s" % (index + 1, e.message))
+        return [rows, errors]
+
+    def process_row(self, row):
+        article = self.get_article_from_row(ARTICLE_COLUMN_FIELD_MAPPING, row)
```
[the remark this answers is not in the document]
said-moj: Thank you. I went with your first suggestion about `get_thing_from_row` knowing about the mappings rather passing it in.

cla_backend/apps/knowledgebase/utils/csv_user_import.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def get_article_category_matrices_from_row(self, article, row):
+        matrices = []
+        for fields in ARTICLE_CATEGORY_MATRIX_COLUMN_FIELD_MAPPING:
+            matrix = self.get_get_article_category_matrix_from_row(article, fields, row)
+            if matrix:
+                matrices.append(matrix)
+        return matrices
+
+    def get_get_article_category_matrix_from_row(self, article, fields, row):
```
exonian, line 126: This method name too

cla_backend/apps/knowledgebase/tests/test_csv_user_import.py
```diff
@@ -0,0 +1,199 @@
...
+                "article": article,
+                "telephone_numbers": telephone_numbers,
+                "article_category_matrices": article_category_matrices,
+            }
+        ]
+
+        def save_article_category_matrices(*args, **kwargs):
+            self.assertEqual(Article.objects.count(), 1)
+            self.assertEqual(TelephoneNumber.objects.count(), 2)
+            raise ValueError("Cannot save article category matrices")
```
exonian, line 119: I think I understand why this is mocked, and I like that it allows us to assert the article and phone number counts mid-transaction, but is it possible to let the exception come from the original code at all? ie is it possible to call `super()`?

exonian (commented):
Couple more comments, but this is looking really good

cla_backend/apps/knowledgebase/tests/test_csv_user_import.py
```diff
@@ -0,0 +1,215 @@
...
+    TELEPHONE_COLUMN_FIELD_MAPPING,
+    ARTICLE_CATEGORY_MATRIX_COLUMN_FIELD_MAPPING,
+)
+from core.tests.mommy_utils import make_recipe
+
+
+class KnowledgebaseCSVImporterTester(TestCase):
+    def setUp(self):
+        self.importer = KnowledgebaseCSVImporter()
+        self.row = self.get_csv_row()
```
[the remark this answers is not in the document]
said-moj: updated with a `_CSV_ROW` class variable that gets copied to an instance variable that is used by all the tests

cla_backend/apps/knowledgebase/tests/test_csv_user_import.py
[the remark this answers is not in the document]
said-moj: I have updated the methods that were modifying `self.row` as part of the expected value to use a separate variable for expected value.

cla_backend/apps/knowledgebase/tests/test_csv_user_import.py
said-moj: I don't think the test `save_article_category_matrices` gets a self argument, so it can't seem to do super.

Instead I have made a copy of the original method and called that after doing asserts:
`        def save_article_category_matrices(article, matrices):
            self.assertEqual(Article.objects.count(), 1)
            self.assertEqual(TelephoneNumber.objects.count(), 2)
            matrices[0] = None
            return self.importer._original_save_article_category_matrices(article, matrices)

        self.importer._original_save_article_category_matrices = self.importer.save_article_category_matrices
        self.importer.save_article_category_matrices = save_article_category_matrices
`

said-moj: Ideally I would have like to have done `rows[0]["article_category_matrices"][0].article = None` inside the `with` context that is expecting the exception but it seems django is doing model not null checking at assignment

exonian: Ah yes my mistake, and that looks like a really clever solution to me

exonian (approved):
This looks fantastic to me, superb work

exonian (approved):

PR #696, opened 2021-02-26
LGA 1568 discrimination provider allocation

Pheonnexx:

This allows for automatic provider allocation to happen with discrimination providers.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/cla_provider/helpers.py
+++ b/cla_backend/apps/cla_provider/helpers.py
@@ -205,7 +205,7 @@ class ProviderAllocationHelper(object):
         return self._get_random_provider(category)
 
     def get_suggested_provider(self, category):
-        non_rota_hours = settings.NON_ROTA_OPENING_HOURS[getattr(category, "code", "")]
+        non_rota_hours = settings.NON_ROTA_OPENING_HOURS[getattr(category, "code")]
         if self.as_of not in non_rota_hours:
             return self._get_rota_provider(category)
         if not os.path.isfile("/tmp/DISABLE_BEST_FIT_PROVIDER"):
--- a/cla_backend/apps/cla_provider/helpers.py
+++ b/cla_backend/apps/cla_provider/helpers.py
@@ -205,10 +205,7 @@ class ProviderAllocationHelper(object):
         return self._get_random_provider(category)
 
     def get_suggested_provider(self, category):
-        if category.id == 8:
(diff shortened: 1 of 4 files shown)


cla_backend/apps/cla_provider/helpers.py
```diff
@@ -205,7 +205,11 @@ def _get_best_fit_provider(self, category):
         return self._get_random_provider(category)
 
     def get_suggested_provider(self, category):
-        if self.as_of not in settings.NON_ROTA_OPENING_HOURS:
+        if category.id == 8:
```
exonian, line 208: Is there a way to make this check not be the id? Is there an appropriately computer-readable field relating to the category name on the model we could use instead?

Ideally, that field would also be used as the key in the `NON_ROTA_OPENING_HOURS` dict so then could be used directly and would scale nicer once we start getting more exceptions to the rule

cla_backend/settings/base.py
```diff
@@ -316,8 +317,12 @@ def env_var_truthy_intention(name):
 CALL_CENTRE_NOTIFY_EMAIL_ADDRESS = os.environ.get("CALL_CENTRE_NOTIFY_EMAIL_ADDRESS", DEFAULT_EMAIL_TO)
 
 NON_ROTA_HOURS = {"weekday": (datetime.time(8, 0), datetime.time(17, 0))}
+DISCRIMINATION_NON_ROTA_HOURS = {"weekday": (datetime.time(8, 0), datetime.time(18, 0))}
 
-NON_ROTA_OPENING_HOURS = OpeningHours(**NON_ROTA_HOURS)
+# Opening hours for a provider of a particular category, the key matches the category key
+NON_ROTA_OPENING_HOURS = defaultdict(
+    lambda: OpeningHours(**NON_ROTA_HOURS), {"discrimination": OpeningHours(**DISCRIMINATION_NON_ROTA_HOURS)}
```
exonian, line 324: This is clever. But it took me a little while to parse. There's a couple of things going on here at once, and in addition this is a defaultdict, which we're not really dealing with all that often so have probably slightly forgotten how they work.

I think maybe, at the expense of conciseness, we could consider doing it in two steps:
1) initialise the defaultdict with the lamba for its default value
2) update the dict with the discrimination key:value pair

exonian (changes requested):
This is fantastic. I think we need to just work on the readability of that defaultdict but it's such a clever solution, and neat and readable in use

exonian (approved):
Fantastic

PR #705, opened 2021-04-22
[WIP] LGA-1121 - Old complaints appearing in complaints tab on CHS

mayowaadeniyi:
- Creates a `FindAndDeleteCasesUsingCreationTime` class
- This is used by a `Command` class. The command does 3 things based on the arguments passed in:
1.  Finds old cases - an old case being a case that is 2 years ago or older and it's most recent log is 2 years old or older
    - If this command is run from a **test suite**, it will return all cases found
    - If this command is run from **cmd line**, it will print out the number of cases found 
2. Deletes cases - deletes all old cases that meet the criteria above
- Subclass the `DeleteOldData` class, so we can make use of the delete functionality 
- Override the `get_eligible_cases` method, so we can change what time `now` refers to (using the `freezetime` library)
- Create test cases based on [these test cases](https://docs.google.com/document/d/1qxKx4l4GCH1PvHSTf8AyP-zwBjWBQq8_EoZCFi4A2LE/edit)

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/cla_butler/tasks.py
+++ b/cla_backend/apps/cla_butler/tasks.py
@@ -64,8 +64,10 @@ class DeleteOldData(Task):
     data = serializers.serialize("json", SomeModel.objects.all())
     """
 
-    def run(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
         self._setup()
+
+    def run(self, *args, **kwargs):
         self.cleanup_cases()
         self.cleanup_diagnosis()
         self.cleanup_eligibility_check()
--- a/cla_backend/apps/cla_butler/tasks.py
+++ b/cla_backend/apps/cla_butler/tasks.py
@@ -112,8 +112,7 @@ class DeleteOldData(Task):
         tokens.delete()
 
     def cleanup_cases(self):
-        two_years = self.now - relativedelta(years=2)
-        cases = Case.objects.filter(modified__lte=two_years)
+        cases = self.get_eligible_cases()
         pks = get_pks(cases)
         from_cases = Case.objects.filter(from_case_id__in=pks)
         fpks = get_pks(from_cases)
@@ -131,6 +130,10 @@ class DeleteOldData(Task):
         self.cleanup_model_from_case(pks, ReasonForContacting)
         self._delete_objects(from_cases)
         self._delete_objects(cases)
+
+    def get_eligible_cases(self):
+        two_years = self.now - relativedelta(years=2)
+        return Case.objects.filter(modified__lte=two_years)
 
     def cleanup_model_from_case(self, pks, model, attr="case_id", case_log_attr=None):
         attr_in = "{attribute}__in".format(attribute=attr)
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -3,25 +3,13 @@ from dateutil.relativedelta import relativedelta
 
 from legalaid.models import Case
 from cla_butler.tasks import DeleteOldData
-from cla_eventlog.models import Log
 
 
 class FindAndDeleteCasesUsingCreationTime(DeleteOldData):
     def get_eligible_cases(self):
         two_years = self.now - relativedelta(years=2)
-        old_cases = []
 
-        filtered_cases = Case.objects.filter(created__lte=two_years)
-        for case in filtered_cases:
-            try:
-                latest_log = Log.objects.filter(case__id=case.id).latest("created")
-                if latest_log.created <= two_years:
-                    old_cases.append(case.id)
-            except Log.DoesNotExist:
-                # Old case without any event logs
-                old_cases.append(case.id)
-
-        return Case.objects.filter(id__in=old_cases)
+        return Case.objects.filter(created__lte=two_years).exclude(log__created__gte=two_years)
 
 
 class Command(BaseCommand):
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -28,17 +28,17 @@ class Command(BaseCommand):
                 instance.run()
             else:
                 return cases
-
-        if django_command == "find_and_delete_old_cases":  # If command is run in terminal
-            if len(args) > 1 and args[1] == "no-input":
-                instance.run()
-            elif args and args[0] == "delete":
-                answer = raw_input(
-                    "Number of cases that will be deleted: {0}\nAre you sure about this? (Yes/No) ".format(
-                        cases.count()
-                    )
-                )
-                if answer == "Yes":
+        else:  # If command is run in terminal
+            if args and args[0] == "delete":
+                if len(args) > 1 and args[1] == "no-input":
                     instance.run()
+                else:
+                    answer = raw_input(
+                        "Number of cases that will be deleted: {0}\nAre you sure about this? (Yes/No) ".format(
+                            cases.count()
+                        )
+                    )
+                    if answer == "Yes":
+                        instance.run()
             else:
                 print("Number of cases to be deleted: " + str(cases.count()))
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -30,7 +30,7 @@ class Command(BaseCommand):
             else:
                 answer = raw_input(
                     "Number of cases that will be deleted: {0}\nAre you sure about this? (Yes/No) ".format(
-                        str(cases.count())
+                        cases.count()
                     )
                 )
                 if answer == "Yes":
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -22,12 +22,17 @@ class Command(BaseCommand):
         instance = FindAndDeleteCasesUsingCreationTime()
         cases = instance.get_eligible_cases()
         django_command = sys.argv[1]
-        if args and args[0] == "delete":
-            if len(args) > 1 and args[1] == "no-input":
-                instance.run()
-            elif django_command == "test":
+
+        if django_command == "test":  # If command is run in test
+            if args and args[0] == "delete":
                 instance.run()
             else:
+                return cases
+
+        if django_command == "find_and_delete_old_cases":  # If command is run in terminal
+            if len(args) > 1 and args[1] == "no-input":
+                instance.run()
+            elif args and args[0] == "delete":
                 answer = raw_input(
                     "Number of cases that will be deleted: {0}\nAre you sure about this? (Yes/No) ".format(
                         cases.count()
@@ -35,7 +40,5 @@ class Command(BaseCommand):
                 )
                 if answer == "Yes":
                     instance.run()
-        elif sys.argv[1] == "test":
-            return cases
-        else:
-            print("Number of cases to be deleted: " + str(cases.count()))
+            else:
+                print("Number of cases to be deleted: " + str(cases.count()))
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -21,16 +21,17 @@ class Command(BaseCommand):
     def handle(self, *args, **kwargs):
         instance = FindAndDeleteCasesUsingCreationTime()
         cases = instance.get_eligible_cases()
-        if len(args) and args[0] == "delete":
+        django_command = sys.argv[1]
+        if args and args[0] == "delete":
             if len(args) > 1 and args[1] == "no-input":
                 instance.run()
-            elif sys.argv[1] == "test":
+            elif django_command == "test":
                 instance.run()
             else:
                 answer = raw_input(
-                    "Number of cases that will be deleted: "
-                    + str(cases.count())
-                    + "\nAre you sure about this? (Yes/No) "
+                    "Number of cases that will be deleted: {0}\nAre you sure about this? (Yes/No) ".format(
+                        str(cases.count())
+                    )
                 )
                 if answer == "Yes":
                     instance.run()
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -22,7 +22,7 @@ class Command(BaseCommand):
         instance = FindAndDeleteCasesUsingCreationTime()
         cases = instance.get_eligible_cases()
         if len(args) == 0:
-            print(cases.count())
+            print("Number of cases to be deleted: " + str(cases.count()))
         elif args[0] == "test_find":
             return cases
         elif args[0] == "delete":
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -14,7 +14,15 @@ class FindAndDeleteCasesUsingCreationTime(DeleteOldData):
 
 
 class Command(BaseCommand):
-    help = "Find and delete cases that are 2 years old or over that were not deleted prior to the task command being fixed"
+    help = (
+        "Find or delete cases that are 2 years old or over that were not deleted prior to the task command being fixed"
+    )
 
     def handle(self, *args, **kwargs):
-        FindAndDeleteCasesUsingCreationTime().run()
+        if args[0] == "find":
+            cases = FindAndDeleteCasesUsingCreationTime().get_eligible_cases()
+            print(cases)
+            print(cases.count())
+
+        if args[0] == "delete":
+            FindAndDeleteCasesUsingCreationTime().run()
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -22,7 +22,18 @@ class Command(BaseCommand):
         instance = FindAndDeleteCasesUsingCreationTime()
         cases = instance.get_eligible_cases()
         if len(args) and args[0] == "delete":
-            instance.run()
+            if len(args) > 1 and args[1] == "no-input":
+                instance.run()
+            elif sys.argv[1] == "test":
+                instance.run()
+            else:
+                answer = raw_input(
+                    "Number of cases that will be deleted: "
+                    + str(cases.count())
+                    + "\nAre you sure about this? (Yes/No) "
+                )
+                if answer == "Yes":
+                    instance.run()
         elif sys.argv[1] == "test":
             return cases
         else:
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -7,7 +7,7 @@ from cla_butler.tasks import DeleteOldData
 
 class FindAndDeleteCasesUsingCreationTime(DeleteOldData):
     def get_eligible_cases(self):
-        super(FindAndDeleteCasesUsingCreationTime, self)._setup()
+        self._setup()
         two_years = self.now - relativedelta(years=2)
 
         return Case.objects.filter(created__lte=two_years).exclude(log__created__gte=two_years)
@@ -19,10 +19,9 @@ class Command(BaseCommand):
     )
 
     def handle(self, *args, **kwargs):
-        if args[0] == "find":
-            cases = FindAndDeleteCasesUsingCreationTime().get_eligible_cases()
-            print(cases)
+        instance = FindAndDeleteCasesUsingCreationTime()
+        if len(args) == 0:
+            cases = instance.get_eligible_cases()
             print(cases.count())
-
-        if args[0] == "delete":
-            FindAndDeleteCasesUsingCreationTime().run()
+        elif args[0] == "delete":
+            instance.run()
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -20,8 +20,10 @@ class Command(BaseCommand):
 
     def handle(self, *args, **kwargs):
         instance = FindAndDeleteCasesUsingCreationTime()
+        cases = instance.get_eligible_cases()
         if len(args) == 0:
-            cases = instance.get_eligible_cases()
             print(cases.count())
+        elif args[0] == "test_find":
+            return cases
         elif args[0] == "delete":
             instance.run()
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -1,3 +1,4 @@
+import sys
 from django.core.management.base import BaseCommand
 from dateutil.relativedelta import relativedelta
 
@@ -7,7 +8,6 @@ from cla_butler.tasks import DeleteOldData
 
 class FindAndDeleteCasesUsingCreationTime(DeleteOldData):
     def get_eligible_cases(self):
-        self._setup()
         two_years = self.now - relativedelta(years=2)
 
         return Case.objects.filter(created__lte=two_years).exclude(log__created__gte=two_years)
@@ -21,9 +21,9 @@ class Command(BaseCommand):
     def handle(self, *args, **kwargs):
         instance = FindAndDeleteCasesUsingCreationTime()
         cases = instance.get_eligible_cases()
-        if len(args) == 0:
-            print("Number of cases to be deleted: " + str(cases.count()))
-        elif args[0] == "test_find":
-            return cases
-        elif args[0] == "delete":
+        if len(args) and args[0] == "delete":
             instance.run()
+        elif sys.argv[1] == "test":
+            return cases
+        else:
+            print("Number of cases to be deleted: " + str(cases.count()))
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -7,6 +7,7 @@ from cla_butler.tasks import DeleteOldData
 
 class FindAndDeleteCasesUsingCreationTime(DeleteOldData):
     def get_eligible_cases(self):
+        super(FindAndDeleteCasesUsingCreationTime, self)._setup()
         two_years = self.now - relativedelta(years=2)
 
         return Case.objects.filter(created__lte=two_years).exclude(log__created__gte=two_years)
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -1,0 +1,31 @@
+from django.core.management.base import BaseCommand
+from dateutil.relativedelta import relativedelta
+
+from legalaid.models import Case
+from cla_butler.tasks import DeleteOldData
+from cla_eventlog.models import Log
+
+
+class FindAndDeleteCasesUsingCreationTime(DeleteOldData):
+    def get_eligible_cases(self):
+        two_years = self.now - relativedelta(years=2)
+        old_cases = []
+
+        filtered_cases = Case.objects.filter(created__lte=two_years)
+        for case in filtered_cases:
+            try:
+                latest_log = Log.objects.filter(case__id=case.id).latest("created")
+                if latest_log.created <= two_years:
+                    old_cases.append(case.id)
+            except Log.DoesNotExist:
+                # Old case without any event logs
+                old_cases.append(case.id)
+
+        return Case.objects.filter(id__in=old_cases)
+
+
+class Command(BaseCommand):
+    help = "Find and delete cases that are 2 years old or over that were not deleted prior to the task command being fixed"
+
+    def handle(self, *args, **kwargs):
+        FindAndDeleteCasesUsingCreationTime().run()
--- a/cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
@@ -48,7 +48,7 @@ class FindAndDeleteOldCases(TestCase):
         freezer = freeze_time(current_time)
         freezer.start()
         # Using handle method because command execute method can only return a string i.e it can't return a queryset
-        cases = self.command.handle("test_find")
+        cases = self.command.handle()
         freezer.stop()
         return cases
 
--- a/cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
@@ -203,6 +203,7 @@ class FindAndDeleteOldCases(TestCase):
         case_1 = self.create_case(date, "legalaid.eligible_case")
         case_2 = self.create_case(date, "legalaid.eligible_case")
         self.create_event_log_for_case(case_0, "MT_CHANGED", _make_datetime(year=2014, month=5, day=30, hour=9))
+        self.create_event_log_for_case(case_0, "MT_CHANGED", _make_datetime(year=2018, month=5, day=30, hour=9))
         self.create_event_log_for_case(case_1, "MT_CHANGED", _make_datetime(year=2014, month=5, day=30, hour=9))
         self.create_event_log_for_case(case_2, "MT_CHANGED", _make_datetime(year=2020, month=4, day=27, hour=9))
 
@@ -221,7 +222,8 @@ class FindAndDeleteOldCases(TestCase):
         self.assertEqual(oldCasesFound.count(), 2)
 
         self.delete_old_cases(dt)
-        self.assertEqual(Case.objects.count(), 1)
+        valid_case = Case.objects.filter(id=case_2.id)
+        self.assertEqual(valid_case.count(), 1)
         self.assertEqual(Log.objects.count(), 1)
         self.assertEqual(AuditLog.objects.count(), 1)
         self.assertEqual(Complaint.objects.count(), 1)
--- a/cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
@@ -222,8 +222,9 @@ class FindAndDeleteOldCases(TestCase):
         self.assertEqual(oldCasesFound.count(), 2)
 
         self.delete_old_cases(dt)
-        valid_case = Case.objects.filter(id=case_2.id)
-        self.assertEqual(valid_case.count(), 1)
+        case_ids = Case.objects.values_list("id", flat=True)
+        # Check there is only 1 case and that case is the relatively new case
+        self.assertEqual(list(case_ids), [case_2.id])
         self.assertEqual(Log.objects.count(), 1)
         self.assertEqual(AuditLog.objects.count(), 1)
         self.assertEqual(Complaint.objects.count(), 1)
--- a/cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
@@ -3,6 +3,7 @@ from django.utils import timezone
 from dateutil.relativedelta import relativedelta
 from freezegun import freeze_time
 
+from cla_butler.tasks import get_pks
 from core.tests.mommy_utils import make_recipe
 from cla_eventlog.models import Log
 from legalaid.models import Case
@@ -46,14 +47,26 @@ class FindAndDeleteOldCases(TestCase):
 
         self.assertEqual(len(logs), 9)
 
-        event_logs_created_two_years_ago = Log.objects.filter(
-            # Cases created two years ago or over
-            case__created__lte=timezone.now() + relativedelta(years=-2),
-            # That have logs created two years ago or over
-            created__lte=timezone.now() + relativedelta(years=-2),
-            # Where the code field on the log = "CASE_VIEWED"
-            code__contains="CASE_VIEWED",
-        )
-        self.assertEqual(len(event_logs_created_two_years_ago), 2)
-        for log in event_logs_created_two_years_ago:
-            self.assertIn(log.case.id, [old_case_2.id])
+        case_pks = get_pks(Case.objects.all())
+        cases_and_logs = {}
+
+        for case_pk in case_pks:
+            case = Case.objects.get(id=case_pk)
+            # Logs are arranged in descending order based upon the event_logs Meta ordering setting
+            case_logs = Log.objects.filter(case__id=case_pk, code="CASE_VIEWED")
+            cases_and_logs[case.reference] = case_logs
+
+        cases_to_delete = []
+        two_years_ago = timezone.now() + relativedelta(years=-2)
+
+        for case_reference, case_logs in cases_and_logs.items():
+            latest_log = case_logs.first()
+            if latest_log.created <= two_years_ago:
+                cases_to_delete.append(case_reference)
+
+        self.assertEqual(len(cases_to_delete), 1)
+
+        for case_reference, case_logs in cases_and_logs.items():
+            case_id = Case.objects.get(reference=case_reference).id
+            for log in case_logs:
+                self.assertIn(log.case.id, [case_id])
--- a/cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
@@ -17,6 +17,12 @@ class FindAndDeleteOldCases(TestCase):
         freezer.stop()
         return case
 
+    def create_event_log_for_case(self, case, code, created=None):
+        if created:
+            make_recipe("cla_eventlog.log", case=case, code=code, created=created)
+        else:
+            make_recipe("cla_eventlog.log", case=case, code=code)
+
     def test_find_cases_viewed_two_years_ago(self):
         new_case = make_recipe("legalaid.case")
         old_case_1 = self.create_three_year_old_case()
@@ -24,18 +30,30 @@ class FindAndDeleteOldCases(TestCase):
 
         self.assertEqual(len(Case.objects.all()), 3)
 
-        make_recipe("cla_eventlog.log", case=new_case, code="CASE_VIEWED")
-        make_recipe("cla_eventlog.log", case=old_case_1, code="CASE_VIEWED")
-        make_recipe(
-            "cla_eventlog.log", case=old_case_2, code="CASE_VIEWED", created=timezone.now() + relativedelta(years=-3)
+        self.create_event_log_for_case(new_case, "CASE_VIEWED")
+        self.create_event_log_for_case(new_case, "CASE_CREATED")
+        self.create_event_log_for_case(new_case, "CASE_VIEWED", created=timezone.now() + relativedelta(years=-3))
+
+        self.create_event_log_for_case(old_case_1, "CASE_VIEWED")
+        self.create_event_log_for_case(old_case_1, "MT_CHANGED")
+        self.create_event_log_for_case(old_case_1, "CASE_VIEWED", created=timezone.now() + relativedelta(years=-1))
+
+        self.create_event_log_for_case(old_case_2, "CASE_VIEWED", created=timezone.now() + relativedelta(years=-3))
+        self.create_event_log_for_case(old_case_2, "CASE_CREATED", created=timezone.now() + relativedelta(years=-3))
+        self.create_event_log_for_case(old_case_2, "CASE_VIEWED", created=timezone.now() + relativedelta(years=-2))
+
+        logs = Log.objects.all()
+
+        self.assertEqual(len(logs), 9)
+
+        event_logs_created_two_years_ago = Log.objects.filter(
+            # Cases created two years ago or over
+            case__created__lte=timezone.now() + relativedelta(years=-2),
+            # That have logs created two years ago or over
+            created__lte=timezone.now() + relativedelta(years=-2),
+            # Where the code field on the log = "CASE_VIEWED"
+            code__contains="CASE_VIEWED",
         )
-
-        self.assertEqual(len(Log.objects.all()), 3)
-
-        logs_for_cases_created_two_years_old = Log.objects.filter(
-            case__created__lte=timezone.now() + relativedelta(years=-2)
-        )
-        case_viewed_logs = logs_for_cases_created_two_years_old.filter(code__contains="CASE_VIEWED")
-        cases_viewed_two_years_ago = case_viewed_logs.filter(created__lte=timezone.now() + relativedelta(years=-2))
-
-        self.assertEqual(len(cases_viewed_two_years_ago), 1)
+        self.assertEqual(len(event_logs_created_two_years_ago), 2)
+        for log in event_logs_created_two_years_ago:
+            self.assertIn(log.case.id, [old_case_2.id])
--- a/cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
@@ -9,7 +9,7 @@ from complaints.models import Complaint
 from legalaid.models import Case, EODDetails, EligibilityCheck, PersonalDetails
 from diagnosis.models import DiagnosisTraversal
 from cla_auditlog.models import AuditLog
-from cla_eventlog.management.commands.find_and_delete_old_cases import FindAndDeleteCasesUsingCreationTime
+from cla_eventlog.management.commands.find_and_delete_old_cases import Command
 
 
 def _make_datetime(year=None, month=None, day=None, hour=0, minute=0, second=0):
@@ -24,7 +24,7 @@ def _make_datetime(year=None, month=None, day=None, hour=0, minute=0, second=0):
 class FindAndDeleteOldCases(TestCase):
     def setUp(self):
         super(FindAndDeleteOldCases, self).setUp()
-        self.instance = FindAndDeleteCasesUsingCreationTime()
+        self.instance = Command()
 
     def create_case(self, current_time, case_type="legalaid.case"):
         case = None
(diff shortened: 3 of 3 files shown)


cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,41 @@
...
+            "cla_eventlog.log", case=old_case_2, code="CASE_VIEWED", created=timezone.now() + relativedelta(years=-3)
+        )
+
+        self.assertEqual(len(Log.objects.all()), 3)
+
+        logs_for_cases_created_two_years_old = Log.objects.filter(
+            case__created__lte=timezone.now() + relativedelta(years=-2)
+        )
+        case_viewed_logs = logs_for_cases_created_two_years_old.filter(code__contains="CASE_VIEWED")
+        cases_viewed_two_years_ago = case_viewed_logs.filter(created__lte=timezone.now() + relativedelta(years=-2))
```
exonian, line 39: Your variable naming has changed from logs to cases. Why?

exonian (commented):
You need to add some more logs to the cases and see how that affects what you have. Right now you have exactly one log per case, so effectively you're looking at squares and claiming to know how rectangles behave.

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Changed in next commit

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,59 @@
...
+        self.create_event_log_for_case(old_case_2, "CASE_CREATED", created=timezone.now() + relativedelta(years=-3))
+        self.create_event_log_for_case(old_case_2, "CASE_VIEWED", created=timezone.now() + relativedelta(years=-2))
+
+        logs = Log.objects.all()
+
+        self.assertEqual(len(logs), 9)
+
+        event_logs_created_two_years_ago = Log.objects.filter(
+            # Cases created two years ago or over
+            case__created__lte=timezone.now() + relativedelta(years=-2),
```
exonian, line 51: Isn't this redundant given the rest of the filter?

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,59 @@
...
+            # That have logs created two years ago or over
+            created__lte=timezone.now() + relativedelta(years=-2),
+            # Where the code field on the log = "CASE_VIEWED"
+            code__contains="CASE_VIEWED",
```
exonian, line 55: Is this the only type of interaction we care about?

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,59 @@
...
+
+        logs = Log.objects.all()
+
+        self.assertEqual(len(logs), 9)
+
+        event_logs_created_two_years_ago = Log.objects.filter(
+            # Cases created two years ago or over
+            case__created__lte=timezone.now() + relativedelta(years=-2),
+            # That have logs created two years ago or over
+            created__lte=timezone.now() + relativedelta(years=-2),
```
exonian, line 53: This tells us about cases that have logs older than two years, but nothing about whether they have logs newer than two years ago, right?
ie if a case was viewed three years ago and also last week, you would return the three year old log

exonian (commented):
Some nice bits but lots to work on.
In particular, have a think about (or try it, and see what happens!) what happens if you add a `CASE_VIEWED` log for `old_case_2` and whether that's desirable. And, if it's not, how your approach might need to change.

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Yeah good point

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Yeah good point, I'll change this now

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: This captures all entry to the case. So I'm not sure what else we would want to look for

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,72 @@
...
+        self.create_event_log_for_case(old_case_2, "CASE_VIEWED", created=timezone.now() + relativedelta(years=-3))
+        self.create_event_log_for_case(old_case_2, "CASE_CREATED", created=timezone.now() + relativedelta(years=-3))
+        self.create_event_log_for_case(old_case_2, "CASE_VIEWED", created=timezone.now() + relativedelta(years=-2))
+
+        logs = Log.objects.all()
+
+        self.assertEqual(len(logs), 9)
+
+        case_pks = get_pks(Case.objects.all())
+        cases_and_logs = {}
```
exonian, line 51: Why do we need a dict? What part of your code does it facilitate?

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,72 @@
...
+
+        for case_pk in case_pks:
+            case = Case.objects.get(id=case_pk)
```
exonian, line 54: How many SQL queries are we going to end up running, given that we have 184,000 `Case`s?
Will that be a problem and, if so, what else can we do?

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,72 @@
...
+
+        self.create_event_log_for_case(old_case_2, "CASE_VIEWED", created=timezone.now() + relativedelta(years=-3))
+        self.create_event_log_for_case(old_case_2, "CASE_CREATED", created=timezone.now() + relativedelta(years=-3))
+        self.create_event_log_for_case(old_case_2, "CASE_VIEWED", created=timezone.now() + relativedelta(years=-2))
+
+        logs = Log.objects.all()
+
+        self.assertEqual(len(logs), 9)
+
+        case_pks = get_pks(Case.objects.all())
```
exonian, line 50: You should put your code in a function somewhere and then run that function in the tests. It shouldn't live inside a test.

exonian (commented):
Take a look at my comment about SQL queries.

Also think about how your code would work if we no longer have any old logs for these case (which may well be the case if we tried to delete them but it failed: we may have deleted their logs but not them). Would a 5-year old case that hasn't been viewed for 3 years and has no logs be caught?

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
exonian: Yeah, maybe. I'm inclined to be more defensive to be honest. I've asked @said-moj for his thoughts on this

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: It's so I can store the all the logs for each case as cases and logs have no direct relationship.

Did it this way to fix the previous issue of finding logs that were over 2 years ago, but not knowing whether the same case had any newer logs.

exonian: Fair enough. But you only write to it once per case and you consume it by running through it as a list of tuples (using .items()) so it could just be a list of tuples. Anyway this may well be moot if you end up restructuring at all

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Hmm yeah it would make it quite slow..

Since every log has reference to a case, I can refactor it, so we iterate over all the logs instead meaning we won't have to make any further SQL queries ?

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Uh not sure what you mean since you highlighted a function I imported. Are you saying I should extract my code into helper functions ?

exonian: I had to comment somewhere!
And I'm not sure, given you refer to it as a helper function we might be talking about different things.i just mean your code, that you're writing. It's not for the test, it's for running to clean up data, so it shouldn't be in the test. It should be in a function, somewhere else

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Ah ok. Most likely I won't be restructuring, rather just reusing the data already there. I'll change it into a tuple

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Oh 😂 ok yeah I understand. I'll move them out. I'll be doing some pairing with Said tomorrow, hopefully PRing will be more graceful 😅

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
exonian: I meant restructuring your code. You may end up not needing an intermediary data structure at all in the end, but while you do I would use the simplest that does what you need

mayowaadeniyi: Uh I edited my comment, but commented too late.. but cool. When I move this into a function, I'll be able to restructure there

exonian: > Yeah since I write to it, I can't make it into a tuple. Unless you mean to convert it afterwards, then sure.
I didn't understand what you meant but I do now. Not a tuple. A list of tuples. The list will be mutable, the inner ones don't need to be.

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Changed it so now we try to find the latest event log rather than specifically look for a 'CASE_VIEWED' log

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Oh yikes I read that wrong. We've since changed it, so it no longer uses a dict. Instead we filter through cases that are less than 2 years old, then filter through the logs based on the cases.

exonian:
I am concerned about the number of SQL queries you're going to be doing.
Please add a test using `assertNumQueries()` (there's an example of that in a test in this repo) or some other technique so that we can understand how many are being done.

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -0,0 +1,31 @@
...
+
+        filtered_cases = Case.objects.filter(created__lte=two_years)
+        for case in filtered_cases:
+            try:
+                latest_log = Log.objects.filter(case__id=case.id).latest("created")
+                if latest_log.created <= two_years:
+                    old_cases.append(case.id)
+            except Log.DoesNotExist:
+                # Old case without any event logs
+                old_cases.append(case.id)
```
exonian, line 22: Could this be combined into the filtering of the `Case` queryset?

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,152 @@
...
+        date = _make_datetime(year=2014, month=4, day=27, hour=9)
+        case = self.create_case(date)
+
+        self.create_event_log_for_case(case, "CASE_VIEWED", _make_datetime(year=2014, month=5, day=30, hour=9))
+        self.create_event_log_for_case(case, "CASE_VIEWED", _make_datetime(year=2018, month=4, day=27, hour=9))
+        self.create_event_log_for_case(case, "CASE_VIEWED", _make_datetime(year=2020, month=4, day=27, hour=9))
+
+        self.delete_old_cases(_make_datetime(year=2021, month=1, day=1, hour=9))
+
+        self.assertEqual(Case.objects.all().count(), 1)
```
exonian, line 52: You don't have to use `.all()` if you're going to be chaining something like `.count()` on

It's enough for you to know at this point that that's generally the rule, and that it applies to everything built-in, like `.count()`, `.filter()`, `.exclude()`, `.first()`, `.update()` etc etc

The technical reason is that all of those are defined on the queryset class (which is what `.all()` returns) but are also deliberately made available on the manager class too (which is what `objects` is), specifically so that they can be chained but don't _have_ to be.

So in all cases where you have `objects.all().count()` you can simplify this to `objects.count()`; it won't actually have any effect on the SQL that is run, because it will already be being processed to the exact same thing.

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: No because we need access to logs and there is no direct relationship between log and case

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Roger that

mayowaadeniyi:
> I am concerned about the number of SQL queries you're going to be doing.
> Please add a test using `assertNumQueries()` (there's an example of that in a test in this repo) or some other technique so that we can understand how many are being done.

In response to this, the `delete_old_cases` method will run 79 queries

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
exonian: The Django ORM lets you filter across ManyToMany relationships

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
exonian: As I say above, strictly speaking no. But all of the ones provided by Django (as opposed to added by us or some third-party library), yes

exonian:
> > I am concerned about the number of SQL queries you're going to be doing.
> > Please add a test using `assertNumQueries()` (there's an example of that in a test in this repo) or some other technique so that we can understand how many are being done.
> 
> In response to this, the `delete_old_cases` method will run 79 queries

On what dataset? Prod? That doesn't seem believable.

exonian:
> Finds old logs - an old log being a log that is 2 years ago or older and/or it's most recent log is 2 years old or older
Deletes logs - deletes all old logs

Is this accurate? You need to be precise, or you'll confuse and mislead future developers.

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -0,0 +1,28 @@
+from django.core.management.base import BaseCommand
+from dateutil.relativedelta import relativedelta
+
+from legalaid.models import Case
+from cla_butler.tasks import DeleteOldData
+
+
+class FindAndDeleteCasesUsingCreationTime(DeleteOldData):
+    def get_eligible_cases(self):
+        super(FindAndDeleteCasesUsingCreationTime, self)._setup()
```
exonian, line 10: I don't think you want to be calling `_setup()` via `super` here.
`super` is generally used for calling the version of the method you're in, from a parent.
eg if we wanted to run the parent's version of `get_eligible_cases()` but then filter the result, we would use `super()` to let us call that parent version of `get_eligible_cases()` from inside `get_eligible_cases()`.
If we want to just call another method on the class, _even if it was defined by a parent class_, we can just call it. `FindAndDeleteCasesUsingCreationTime` will have all the same methods on it as `DeleteOldData`, we don't generally have to worry (or even know!) which were defined on a parent.

There _are_ reasons to call a different method with `super`, but they're not usual.

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -0,0 +1,28 @@
...
+
+class Command(BaseCommand):
+    help = (
+        "Find or delete cases that are 2 years old or over that were not deleted prior to the task command being fixed"
+    )
+
+    def handle(self, *args, **kwargs):
+        if args[0] == "find":
+            cases = FindAndDeleteCasesUsingCreationTime().get_eligible_cases()
+            print(cases)
```
exonian, line 24: Do you really want to print thousands of cases to the stdout?

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -0,0 +1,28 @@
...
+
+        return Case.objects.filter(created__lte=two_years).exclude(log__created__gte=two_years)
+
+
+class Command(BaseCommand):
+    help = (
+        "Find or delete cases that are 2 years old or over that were not deleted prior to the task command being fixed"
+    )
+
+    def handle(self, *args, **kwargs):
```
exonian, line 21: What's the default behaviour if I don't pass either "find" or "delete"? Is this desirable? What *should* the default behaviour be?

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,192 @@
...
+        freezer.start()
+        case = make_recipe(case_type)
+        freezer.stop()
+        return case
+
+    def create_event_log_for_case(self, case, code, created):
+        make_recipe("cla_eventlog.log", case=case, code=code, created=created)
+
+    def delete_old_cases(self, current_time):
+        self.instance.run()
```
exonian, line 41: Why is find_old_cases freezegunned but this isn't?

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,192 @@
...
+
+        self.create_event_log_for_case(case, "CASE_VIEWED", _make_datetime(year=2014, month=5, day=30, hour=9))
+        self.create_event_log_for_case(case, "CASE_VIEWED", _make_datetime(year=2018, month=4, day=27, hour=9))
+        self.create_event_log_for_case(case, "CASE_VIEWED", _make_datetime(year=2020, month=4, day=27, hour=9))
+
+        dt = _make_datetime(year=2021, month=1, day=1, hour=9)
+        oldCasesFound = self.find_old_cases(dt)
+        self.assertEqual(oldCasesFound.count(), 0)
+        # Only need to check for this once as every test case uses this method via calling self.delete_old_cases
+        with self.assertNumQueries(1):
```
exonian, line 63: Given the changes you've made to the management command, I'd like to see an `assertNumQueries` for running the whole command in "find" mode, and not just callings its `get_eligible_cases` method directly

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,192 @@
...
+        date = _make_datetime(year=2014, month=4, day=27, hour=9)
+        case = self.create_case(date)
+
+        self.create_event_log_for_case(case, "CASE_VIEWED", _make_datetime(year=2014, month=5, day=30, hour=9))
+        self.create_event_log_for_case(case, "CASE_VIEWED", _make_datetime(year=2018, month=4, day=27, hour=9))
+        self.create_event_log_for_case(case, "CASE_VIEWED", _make_datetime(year=2020, month=4, day=27, hour=9))
+
+        dt = _make_datetime(year=2021, month=1, day=1, hour=9)
+        oldCasesFound = self.find_old_cases(dt)
+        self.assertEqual(oldCasesFound.count(), 0)
```
exonian, line 61: This is a really bad example to be testing `assertNumQueries` on. From our previous conversations, can you work out why?

exonian (changes requested):
A few things to improve and a couple of misunderstandings.

Now that you've got the management command being able to run just `find`, I don't actually think the thing I had to do to let us call `get_eligible_cases` directly in testing is necessary or desirable any more: I would make your `delete_old_cases` and `find_old_cases` both call the management command (using `call_command` from Django) directly, with the relevant args. This will be a much truer reflection of reality

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: Yeah true. Forgot we have like 100k+ cases to go through

I remember you talking about transactions in the database. Will that act as a cache, so we can restore/rollback the cases if anything goes wrong? (I had this in mind, but didn't know how to achieve it).

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: Thanks, I've made the default `find` since that is what we would want to do first before any deletions.

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: So we always find before doing a delete and both the find and delete use the same instance. I can put it in if it makes it more clearer

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Got it

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Because there wouldn't be any queries right?

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: Thanks again I forgot I could use `self` lol

mayowaadeniyi:
I do have a question about the find command. I can't return a queryset from it as from the docs, it can only return a string. How should I go about using it in the tests?

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
exonian: I think you would need to explicitly invoke a transaction. I recommend searching for "atomic" in the Django docs. Make sure to be on the right Django version, though: that stuff will have evolved various helpers that we won't have access to on our old version

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
exonian: Perfect

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
exonian: That would still run one query (which is what you assert). Finding out it is zero will still take a query.

But try to remember why we were concerned that we might end up with more than one query. What kinds of things did we expect the number of queries to increase with?

exonian:
> I do have a question about the find command. I can't return a queryset from it as from the docs, it can only return a string. How should I go about using it in the tests?

Oh yes, good catch! You're right!
You *might* be able to call `.handle()` directly? If not, you might want to move everything you're doing into methods (still on your command class) that you can then call directly in your testing, and make `.handle()` call them when it's run. Making as much of your code testable as possible, really. If you can call `handle()`, that would be my preference

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Expected it to increase /w number of cases. Ahh ok so it's not about the finding of the cases, but rather making sure only 1 query is run regardless of the number of cases?

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: Right I checked it out and every transaction is committed to the database by default. So we can just undo the transaction from the database. Awesome

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
exonian: Exactly! It *could* increase with the number of cases that exist, but it could also increase with the number of cases that we *return*. So I'd like to see us assert that only 1 query runs even when we find, say, 3 out of 6 cases or something like that

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,196 @@
...
+    month = month if month else today.month
+    day = day if day else today.day
+    dt = datetime.datetime(year, month, day, hour, minute, second)
+    return timezone.make_aware(dt, timezone.get_current_timezone())
+
+
+class FindAndDeleteOldCases(TestCase):
+    def setUp(self):
+        super(FindAndDeleteOldCases, self).setUp()
+        self.instance = Command()
```
exonian, line 27: Could you rename this variable, please?
Because I'd been reading the command previously, when reading your testI was convinced `instance` was the task...

I think `self.command` would be clearest all around

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,196 @@
...
+        freezer.start()
+        case = make_recipe(case_type)
+        freezer.stop()
+        return case
+
+    def create_event_log_for_case(self, case, code, created):
+        make_recipe("cla_eventlog.log", case=case, code=code, created=created)
+
+    def delete_old_cases(self, current_time):
+        self.instance.execute("delete")
```
exonian, line 41: Why does this not freeze time but just finding cases does?

exonian (changes requested):

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -0,0 +1,29 @@
+from django.core.management.base import BaseCommand
+from dateutil.relativedelta import relativedelta
+
+from legalaid.models import Case
+from cla_butler.tasks import DeleteOldData
+
+
+class FindAndDeleteCasesUsingCreationTime(DeleteOldData):
+    def get_eligible_cases(self):
+        self._setup()
```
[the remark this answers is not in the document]
mayowaadeniyi: Hmm I was thinking having out of sync times could mean we potentially delete a case, that wasn't included in the find results. 

I need to run `_setup()` in order for `self.now` to be set though. I was thinking to pass make a condition in the `run` command in tasks that if we included an argument, it wouldn't run `setup`.

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def find_old_cases(self, current_time):
+        cases = None
+        freezer = freeze_time(current_time)
+        freezer.start()
+        # Using handle method because command execute method can only return a string i.e it can't return a queryset
```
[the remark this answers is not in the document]
mayowaadeniyi: It won't work, but I thought that's fine since this is just for having it run in the tests

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Yep

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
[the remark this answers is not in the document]
exonian: Will it break in a confusing way?

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: So we always do a find before doing a delete and both the find and delete use the same instance. I can put it in if it makes it more clearer

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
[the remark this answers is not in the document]
mayowaadeniyi: We are using `handle` here, so it's just calling the function and not executing it like a command. I don't think that should break for any reason?

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
exonian: > and it is only set by the delete command

I don't undersand. Where's it set? You've utterly lost me

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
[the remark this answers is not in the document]
mayowaadeniyi: I was thinking of `setUp`. But then I thought we need to call it again when we use it in the cmd line. Hmm

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Yikes idk what I'm saying. I just got it. I'll do that now

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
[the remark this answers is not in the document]
mayowaadeniyi: Can I put it in Command `__init__`? Will run once regardless of who calls it

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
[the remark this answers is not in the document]
exonian: I mean if it were to be run on the command line

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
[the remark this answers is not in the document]
exonian: Not the command's `__init__`, but the task's `__init__` yep!

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
exonian: Can you link me to what you're looking at, please?

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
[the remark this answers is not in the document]
mayowaadeniyi: Thanks!

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: https://django.readthedocs.io/en/1.7.x/topics/db/transactions.html#django-s-default-transaction-behavior

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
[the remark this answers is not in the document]
mayowaadeniyi: This would break in command line. Is that a problem? Is the issue if we were to run this by accident otherwise I am not sure what's wrong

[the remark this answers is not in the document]
exonian: Just generally good practice not to build things that would break if used in a particular way. It's not a 100% deal-breaker if it turns out to be unreasonably difficult, but I think you should be able to identify whether it's running from the command line or not.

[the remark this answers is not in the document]
mayowaadeniyi: Done

mayowaadeniyi:
When running the delete command from cmd line, shall I make the cmd line wait for input (Y/N) as an extra security check?

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
exonian: Right, so yes transactions are automatically committed when they complete.
And every query is wrapped in a transaction if it's not explicitly part of a bigger transaction.
So if you execute a query, it will run inside a transaction and either be committed or rolled back, and then the transaction is done and gone.

You can also put multiple queries into one transaction, so that they are either all committed or all rolled back together. But you have to actively use something like `atomic` to make these queries share a transaction, and I don't think we are doing that here.

And, either way, I think "So we can just undo the transaction from the database" suggests you might be misunderstanding this a bit. Once a transaction is committed, it's committed. You can roll back a transaction from within it, *instead* of committing it, so we could use one to avoid repeating what happened before, where the `Log`s were deleted but the `Case` failed to delete and we were left with a `Case` with no `Log`s, by wrapping those two deletions in a transaction, but we couldn't use it to restore data after the fact, in case that's what you were thinking.

exonian:
> When running the delete command from cmd line, shall I make the cmd line wait for input (Y/N) as an extra security check?

I think that's potentially a good idea, but look at Django's `collectstatic` command to see what `--noinput` is and decide whether we want to have something like that too

cla_backend/apps/cla_butler/tasks.py
```diff
@@ -64,8 +64,10 @@ class DeleteOldData(Task):
     data = serializers.serialize("json", SomeModel.objects.all())
     """
 
-    def run(self, *args, **kwargs):
+    def __init__(self, *args, **kwargs):
```
exonian, line 67: Perfect

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,199 @@
...
+        make_recipe("cla_eventlog.log", case=case, code=code, created=created)
+
+    def delete_old_cases(self, current_time):
+        freezer = freeze_time(current_time)
+        freezer.start()
+        self.command.execute("delete")
+        freezer.stop()
+
+    def find_old_cases(self, current_time):
+        cases = None
```
exonian, line 47: I don't think you need to set this as you always override it a few lines later

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,199 @@
...
+        self.create_case(date)
+        self.create_case(date)
+        self.create_case(date)
+
+        dt = _make_datetime(year=2021, month=1, day=1, hour=9)
+        oldCasesFound = self.find_old_cases(dt)
+        self.assertEqual(oldCasesFound.count(), 3)
+
+        # Only need to check for this once as every test case uses this method via calling self.find_old_cases
+        with self.assertNumQueries(1):
```
exonian, line 149: Good. This is improving. However, I would put this assertion around the `self.find_old_cases(dt)` as well as the `count()` on the qs (ie the `assertNumQueries` comes before you run the command, and encapsulates both getting the qs and forcing it to evaluate).
We want to be nice and thorough, could easily miss something otherwise, like if the management command did more than return an unevaluated query. We want to know how many queries are run in total

exonian (changes requested):
Starting to look very good.

Couple of comments around transactions and the query number assertion.

Also, I don't like that the result of all of your tests appears to be either keeping all of the cases or deleting all of the cases.

I could write a ridiculous bit of logic like the following for the deletion step and I don't think it would fail any of your tests... 🤔 🙈 

```
if self.get_eligible_cases():
    Case.objects.all().delete()
```

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: Oh ok. So if we wanted to e.g. restore deleted cases can we rollback the database? I was mostly concerned with what we will do in that kinda situation

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Ty this isn't like javascript 😅

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Cool done that

mayowaadeniyi:
> Starting to look very good.
> 
> Couple of comments around transactions and the query number assertion.
> 
> Also, I don't like that the result of all of your tests appears to be either keeping all of the cases or deleting all of the cases.
> 
> I could write a ridiculous bit of logic like the following for the deletion step and I don't think it would fail any of your tests... 🤔 🙈
> 
> ```
> if self.get_eligible_cases():
>     Case.objects.all().delete()
> ```

Yikes >_<. In response to this, I was going to create more cases for each test case, so we know it's not just blindly deleting everything.

Also, for cases that are not going to be deleted (i.e. new cases) we are only making assertions on `Logs` and `Cases`. We only involve things like `Complaints` and `AuditLogs` when the test `case` is eligible for deletion (i.e. an old case). To make make the tests more foolproof, shall I check for more fields e.g. `Complaints` on ALL the test cases instead of just the old test cases ?

mayowaadeniyi:
> > When running the delete command from cmd line, shall I make the cmd line wait for input (Y/N) as an extra security check?
> 
> I think that's potentially a good idea, but look at Django's `collectstatic` command to see what `--noinput` is and decide whether we want to have something like that too

If we were to automate the deletion somehow I think it would be good to have this. So I think it could be good for that reason

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
exonian: Have you? I'm not seeing it!

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
exonian: Not particularly easily, no. We just need to be confident in what we're deleting

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -22,7 +22,18 @@ def handle(self, *args, **kwargs):
         instance = FindAndDeleteCasesUsingCreationTime()
         cases = instance.get_eligible_cases()
         if len(args) and args[0] == "delete":
```
exonian, line 24: `len(args)` and `args` will have the same truthiness, so no need to get the length here purely to check if it's non-zero, we can do `if args and args[0] == "delete"`

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -22,7 +22,18 @@ def handle(self, *args, **kwargs):
...
-            instance.run()
+            if len(args) > 1 and args[1] == "no-input":
+                instance.run()
+            elif sys.argv[1] == "test":
```
exonian, line 27: This is hard to notice (that it's different from args). I would consider making `sys.argv[1]` into a variable with a clear name a bit higher up and then we can use it here without confusion

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -22,7 +22,18 @@ def handle(self, *args, **kwargs):
...
+                instance.run()
+            else:
+                answer = raw_input(
+                    "Number of cases that will be deleted: "
+                    + str(cases.count())
```
exonian, line 32: I would recommend using string formatting, ie something like `"My {} goes here".format("text")`

exonian (changes requested):
Great work, nearly there

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: Cool thanks for clearing that up 👍

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: Done

mayowaadeniyi:
Shall I test on the fields e.g. like `Complaints` and `AuditLogs` on every test case to ensure deletions are correct? Seeing as I need to add more cases/logs to each test case to make highlight it's not just blindly doing 
```
if self.get_eligible_cases():
    Case.objects.all().delete()
```

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -21,16 +21,17 @@ class Command(BaseCommand):
...
-            elif sys.argv[1] == "test":
+            elif django_command == "test":
                 instance.run()
             else:
                 answer = raw_input(
-                    "Number of cases that will be deleted: "
-                    + str(cases.count())
-                    + "\nAre you sure about this? (Yes/No) "
+                    "Number of cases that will be deleted: {0}\nAre you sure about this? (Yes/No) ".format(
+                        str(cases.count())
```
exonian, line 33: You shouldn't need to call `str` on it, `format()` should take care of that for you. I think

exonian:
> Shall I test for every field e.g. like `Complaints` and `AuditLogs` on every test case to ensure deletions are correct? Seeing as I need to add more cases/logs to each test case to make highlight it's not just blindly doing
> 
> ```
> if self.get_eligible_cases():
>     Case.objects.all().delete()
> ```
> 
> Currently new case tests do not check for `Complaints` and `AuditLogs`, only old cases that are due to be deleted.

Not on all, but maybe on one counter-example just to be sure

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: Done

mayowaadeniyi:
Shall I add assertions before and after the delete like Heather did so we can see what got deleted?

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -0,0 +1,233 @@
...
+        make_recipe("complaints.complaint", eod=eod_1, audit_log=[audit_log_2])
+        make_recipe("complaints.complaint", eod=eod_2, audit_log=[audit_log_3])
+
+        dt = _make_datetime(year=2021, month=1, day=1, hour=9)
+        oldCasesFound = self.find_old_cases(dt)
+        self.assertEqual(oldCasesFound.count(), 2)
+
+        self.delete_old_cases(dt)
+        valid_case = Case.objects.filter(id=case_2.id)
+        self.assertEqual(valid_case.count(), 1)
```
exonian, line 226: I would also assert that the total number of cases is 1, but I like that you are asserting it's the right one. But this assertion alone would pass if both cases still existed.

Ooh, even better, you could do something like
`self.assertEqual(Case.objects.values_list('id', flat=True), [case_2.id])`

This will do both things at once!

exonian (commented):
Just one more thing from me.

And per your question, no I don't think you really need to assert the object counts before deletion as you've created them with `modelmommy` within the tests so we can be confident we started with the data we think we did.

And good work fixing the "logs" to "cases" in the PR description, but you've only half done it! You're still talking about logs in step 2, for deletion...

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Ahh yeah true. Awesome, I change `Case.objects.values_list('id', flat=True)` to `list(Case.objects.values_list('id', flat=True))` as the first one is a queryset.

exonian: That's really great work

exonian (commented):
Fantastic! 🥳 

Let's have @said-moj take a look and give it the formal Approve if he thinks it's good. I think it'd benefit from the fresh pair of eyes as I've been working pretty closely on it now

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -0,0 +1,41 @@
...
+                    "Number of cases that will be deleted: {0}\nAre you sure about this? (Yes/No) ".format(
+                        cases.count()
+                    )
+                )
+                if answer == "Yes":
+                    instance.run()
+        elif sys.argv[1] == "test":
+            return cases
+        else:
+            print("Number of cases to be deleted: " + str(cases.count()))
```
said-moj, line 41:
```        django_command = sys.argv[1]
        if args and args[0] == "delete":
            if len(args) > 1 and args[1] == "no-input":
                instance.run()
            elif django_command == "test":
                instance.run()
            else:
                answer = raw_input(
                    "Number of cases that will be deleted: {0}\nAre you sure about this? (Yes/No) ".format(
                        cases.count()
                    )
                )
                if answer == "Yes":
                    instance.run()
        elif sys.argv[1] == "test":
            return cases
        else:
            print("Number of cases to be deleted: " + str(cases.count()))```
  
I find this quite hard to follow, is there a way to simplify by doing something like 
if is_report_operation then
  show report
if is_delete_operation then
   delete cases

If you do that then you don't need to make any exception for tests, as they could mock the `is_delete_operation` or pass in the correct arguments

**Comment by said-moj** [comment]
Is there a test to ensure that a relatively new case with no event logs doesn't get deleted?

**Comment by exonian** [comment]
> Is there a test case to ensure that a relatively new case with no event logs doesn't get deleted?

https://github.com/ministryofjustice/cla_backend/pull/705/files#diff-3ef352cb8e035742d29aebce280fcaa84573f1ae4b00efd6d9929d031eb1e587R124-R134

File: cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py (hunk 28 above)
Comment (mayowaadeniyi) [reply cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py:41]: E.g. for the `is_report_operation`, you can only return a string when running `command.execute()`, so I have to make conditions depending on if it's being run in terminal or via a test i.e. we want to return queryset when run in a test and can only print the number of cases in terminal

I was thinking I could refactor in such a way where I can make the first conditions based upon the environment it's in 🤔

File: cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py [hunk 29]
```diff
@@ -0,0 +1,41 @@
...
+                instance.run()
+            else:
+                answer = raw_input(
+                    "Number of cases that will be deleted: {0}\nAre you sure about this? (Yes/No) ".format(
+                        cases.count()
+                    )
+                )
+                if answer == "Yes":
+                    instance.run()
+        elif sys.argv[1] == "test":
```
exonian, line 38: This should use the `django_command` variable

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
exonian: Yeah I think it could maybe be more readable, but fundamentally it's complex.

The routes we have are:
- cli no args -> find cases and print the count
- cli "delete" arg -> find cases, prompt for confirmation, delete the cases
- cli "delete noinput" args -> find cases, delete the cases
- via test command, no args -> find cases and return queryset
- via test command, "delete" arg -> find cases, delete the cases

So yeah I agree with the direction @said-moj's going, but maybe not quite with some of the conclusions about the testing thing (I think maybe I'm erring on the side of avoiding mocking the actual moving part that should be under test? I'm not sure).

You could add `find()` and `find_and_delete()` methods to the management command (alongside `handle()`) and hand off to those with your `if args and args[0] == "delete":`

So it'd look something like
```
if args and args[0] == "delete":
    return self.find_and_delete()
else:
    return self.find()
```
with all the delete logic you have now inside one method, and the fallback to either return qs or print count inside the other. No less code, still special-casing tests behaviour, but potentially much easier to read.

mayowaadeniyi: I've refactored it so the first condition is based on what environment it is run in. Hopefully this is more readable?

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: Refactored it a bit with this in mind

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -22,20 +22,23 @@ def handle(self, *args, **kwargs):
...
-                instance.run()
-            elif django_command == "test":
+
+        if django_command == "test":  # If command is run in test
+            if args and args[0] == "delete":
                 instance.run()
             else:
+                return cases
+
+        if django_command == "find_and_delete_old_cases":  # If command is run in terminal
```
exonian, line 32: I would remove this condition and just have it as the `else` to your previous one.
If we imagine a situation where I wrote a new management command called "do_cleanup_stuff.py" and inside that I did:
```
call_command("find_and_delete_old_cases")
```
this command wouldn't run properly. That's undesirable. But I do like the idea.

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -22,20 +22,23 @@ def handle(self, *args, **kwargs):
...
+            if len(args) > 1 and args[1] == "no-input":
```
exonian, line 33: I think I'd put this inside the `args[0] == "delete"` condition. I don't think I like the idea that "manage.py find_and_delete_old_cases whatsgoingon no-input" runs the deletion... 😅

exonian (commented):
This refactor looks promising. You can probably ignore my suggestions from a few minutes ago, now 😁 and just focus on this version and my comments on it

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: Done (nested branches ftw 😅 )

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: Thanks. Done

exonian (commented):
Nice. @said-moj what do you think?

said-moj (approved):
👍  LGTM

exonian (approved):

PR #706, opened 2021-04-27
LGA 1604 Split cases

Pheonnexx:

Matter type selection was being enforced,  which in turn meant that cases with categories that dont have matter types can be assigned or split.  The validation for enforcing matter types has been removed and the code checked to make sure that it doesnt require them anywhere else.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/call_centre/tests/test_forms.py
+++ b/cla_backend/apps/call_centre/tests/test_forms.py
@@ -55,6 +55,7 @@ class ProviderAllocationFormTestCase(TestCase):
 
         self.assertEqual(case.provider, provider)
         self.assertEqual(Log.objects.count(), 1)
+        self.assertContains(Log.objects.first().notes, "Assigned to ")
 
     @mock.patch("cla_provider.models.timezone.now")
     @mock.patch("cla_provider.helpers.timezone.now")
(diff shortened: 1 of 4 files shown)


mayowaadeniyi (commented):
Do you mean in the summary `Matter type selection was being enforced, which in turn meant that cases with categories that dont have matter types **can't** be assigned or split`

cla_backend/apps/call_centre/tests/test_forms.py
```diff
@@ -246,40 +246,13 @@ def test_save_without_matter_type_both(self, timezone_mock, models_timezone_mock
...
+        self.assertEqual(Log.objects.count(), 0)
+        form.save(user)
 
-        self.assertFalse(form.is_valid())
-        self.assertDictEqual(
-            form.errors,
-            {"__all__": [u"Can't assign to specialist provider without setting matter_type1 and matter_type2"]},
-        )
+        self.assertEqual(case.provider, provider)
+        self.assertEqual(Log.objects.count(), 1)
```
exonian, line 255: Can we test for the log code? Would make it easier to understand what exactly is being tested, even if it doesn't make the test stricter

exonian (approved):

exonian (approved):

PR #708, opened 2021-05-11
LGA 1707 cb1 report scheduling

Pheonnexx:

This pull request is a small redesign of the CB1 MI report.

  - A new command has been created which runs the report for the current sunday to sunday.  This is scheduled via the cronjobs to run on 5am sundays.
  - View code around the display of reports has been refactored to be clearer in its objective.
  - As well as get reports for specific user there is now a route which will give you all the scheduled reports that have been produced.
  - JS has been amended, if the report is scheduled we no longer display a delete link for it.
  - The new route for showing the *scheduled* cb1 MI reports is behind the feature flag SHOW_NEW_CB1 which can be set to true or false.  Note:  I have increased the day limit on the report to 8 days and this is not behind the feature flag.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/reports/api.py
+++ b/cla_backend/apps/reports/api.py
@@ -3,6 +3,8 @@ from rest_framework.authentication import SessionAuthentication
 from rest_framework import generics
 from rest_framework.permissions import IsAdminUser
 
+from core.models import get_web_user
+
 from .serializers import ExportSerializer
 from .models import Export
 
@@ -13,6 +15,10 @@ class ExportListView(generics.ListAPIView, generics.DestroyAPIView):
     permission_classes = (IsAdminUser,)
     authentication_classes = (SessionAuthentication,)
     page_size = 1000
+    scheduled = False
 
     def get_queryset(self):
-        return super(ExportListView, self).get_queryset().filter(user=self.request.user)
+        if self.scheduled:
+            return super(ExportListView, self).get_queryset().filter(user=get_web_user())
+        else:
+            return super(ExportListView, self).get_queryset().filter(user=self.request.user)
--- a/cla_backend/apps/reports/api.py
+++ b/cla_backend/apps/reports/api.py
@@ -19,6 +19,8 @@ class ExportListView(generics.ListAPIView, generics.DestroyAPIView):
 
     def get_queryset(self):
         if self.scheduled:
-            return super(ExportListView, self).get_queryset().filter(user=get_web_user())
+            user = get_web_user()
         else:
-            return super(ExportListView, self).get_queryset().filter(user=self.request.user)
+            user = self.request.user
+
+        return super(ExportListView, self).get_queryset().filter(user=user)
--- a/cla_backend/apps/reports/management/commands/mi_cb1_report.py
+++ b/cla_backend/apps/reports/management/commands/mi_cb1_report.py
@@ -1,7 +1,9 @@
 # coding=utf-8
 import logging
 from django.core.management.base import BaseCommand
-
+from reports.tasks import ExportTask
+from core.models import get_web_user
+from django.views.decorators.csrf import csrf_exempt
 
 logger = logging.getLogger(__name__)
 
@@ -12,10 +14,11 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         self.create_report()
 
-    def create_report():
-        print("stuff goes here")
-
-        # '{"action": "Export", "csrfmiddlewaretoken": "PQk4Pt55CL0NBapx9hSqZTJkSn6tL6TL", "date_from": "08/05/2021", "date_to": "10/05/2021"}'
+    @csrf_exempt
+    def create_report(self):
+        report_data = '{"action": "Export", "csrfmiddlewaretoken": "PQk4Pt55CL0NBapx9hSqZTJkSn6tL6TL", "date_from": "2021-05-08", "date_to": "2021-05-10"}'
 
         # report_data = json_stuff_goes_here
-        #         ExportTask().delay(user_person.pk, filename_of_report, mi_cb1_extract_agilisys, report_data)
+        web_user = get_web_user()
+        filename_of_report = "WEEKLY-REPORT-TEST.csv"
+        ExportTask().delay(web_user.pk, filename_of_report, "MICB1Extract", report_data)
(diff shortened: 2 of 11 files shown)


cla_backend/apps/reports/views.py
```diff
@@ -34,29 +34,29 @@
...
-                success_task().delay(request.user.pk, filename, form_class.__name__, json.dumps(request.POST))
+    form = form_class()
+    if valid_submit(request, form):
+        success_task().delay(request.user.pk, filename, form_class.__name__, json.dumps(request.POST))
 
-                messages.info(
-                    request, u"Your export is being processed. It " u"will show up in the downloads tab " u"shortly."
-                )
+        messages.info(
+            request, u"Your export is being processed. It " u"will show up in the downloads tab " u"shortly."
```
said-moj, line 50: Can this be one string?

cla_backend/apps/reports/management/commands/mi_cb1_report.py
```diff
@@ -0,0 +1,29 @@
...
+
+    def handle(self, *args, **options):
+        self.create_report()
+
+    @csrf_exempt
+    def create_report(self):
+        date_from = (date.today() - timedelta(days=7)).strftime("%Y-%m-%d")
+        date_to = date.today().strftime("%Y-%m-%d")
+
+        report_data = '{{"action": "Export", "csrfmiddlewaretoken": "PQk4Pt55CL0NBapx9hSqZTJkSn6tL6TL", "date_from": "{0}", "date_to": "{1}"}}'.format(
```
said-moj, line 23: Do we need this hardcoded `csrfmiddlewaretoken`?

Pheonnexx: this has been removed.

cla_backend/apps/reports/management/commands/mi_cb1_report.py
```diff
@@ -0,0 +1,27 @@
...
+
+    @csrf_exempt
+    def create_report(self):
+        date_from = (date.today() - timedelta(days=7)).strftime("%Y-%m-%d")
+        date_to = date.today().strftime("%Y-%m-%d")
+
+        report_data = '{{"action": "Export", "date_from": "{0}", "date_to": "{1}"}}'.format(date_from, date_to)
+
+        web_user = get_web_user()
+        filename_of_report = "scheduled-mi-cb1-extract.csv"
```
said-moj, line 26: Does the report get overwritten if you run the report again or some additional processing is added to make the filename unique (i.e adding a date/time suffix)

cla_backend/apps/reports/views.py
Pheonnexx: this has been amended

cla_backend/apps/reports/api.py
```diff
@@ -13,6 +15,10 @@ class ExportListView(generics.ListAPIView, generics.DestroyAPIView):
     permission_classes = (IsAdminUser,)
     authentication_classes = (SessionAuthentication,)
     page_size = 1000
+    scheduled = False
 
     def get_queryset(self):
-        return super(ExportListView, self).get_queryset().filter(user=self.request.user)
+        if self.scheduled:
```
exonian, line 21: Given these lines are otherwise the same, I wonder if it'd be easier on the eye to conditionally set `user` and then be able to just call the `super` in the same way regardless

cla_backend/apps/reports/urls.py
```diff
@@ -7,6 +7,7 @@
 urlpatterns = patterns(
     "",
     url(r"^api/exports/$", api.ExportListView.as_view(), name="exports"),
+    url(r"^api/exports/scheduled/$", api.ExportListView.as_view(scheduled=True), name="exports"),
```
exonian, line 10: So, the `name` here can be used by Django for reverse url lookups ie getting a url from the url pattern name.
As you can see, we already have three url patterns that use the same `name` ("exports").
However, they all have different signatures in terms of url parameters, that is:

- reverse with name = exports -> api/exports/
- reverse with name = exports and a pk kwarg -> api/exports/<pk>/
- reverse with name = exports and a filename -> exports/download/<file_name>/

so depending what `reverse()` is called with, it's unambiguous which url pattern should be used.

However, if we add the scheduled list pattern with the same name, the reverse lookup would now be ambiguous. I think it would default to the first one, and it might not strictly matter for us in this case, but I'd prefer to avoid it with a unique (or, strictly, unique given its signature) `name`

cla_backend/apps/reports/views.py
```diff
@@ -193,31 +201,11 @@ def download_file(request, file_name="", *args, **kwargs):
     response["Content-Disposition"] = "attachment; filename=%s" % smart_str(file_name)
     response["X-Sendfile"] = smart_str("%s%s" % (settings.TEMP_DIR, file_name))
 
-    try:
-        export_record = Export.objects.get(user_id=request.user.pk, path__endswith=file_name)
-        export_record.delete()
-    except Export.DoesNotExist:
-        raise Http404("Export does not exist")
+    if not file_name.__contains__("scheduled"):
```
exonian, line 204: Shouldn't have to use the double underscores on both sides ("dunder") methods directly.
They're there to define the behaviour of python's keyword syntax.

Is there a specific reason this cannot be `if "scheduled" not in file_name`?

exonian, line 204: I was half way through writing "it's not clear to me why we only have to get the file if it's a scheduled one" when I realised what was actually going on.
Granted, I wasn't reading the whole method in order or I might have twigged, but maybe moving the try/except into something like a `self.delete_export()` method might make it super clear what's happening

exonian (commented):
Couple of suggestions.
This is really good work, very tidy and nicely done

cla_backend/apps/reports/management/commands/mi_cb1_report.py
Pheonnexx: In ExportTaskBase() the filename is set with a date attached, so we do not over write reports.

cla_backend/apps/reports/views.py
Pheonnexx: the specific reason would be : brain fog.  Quite literally dont know why i went this route.

exonian (approved):
Looks great!

PR #709, opened 2021-06-09
LGA-1719 - Timezone-aware assigned_out_of_hours calculation

exonian:
Uses the new `tz_aware` argument to `OpeningHours.available()` to calculate the value of `assigned_out_of_hours` in a timezone-aware, and therefore correct, way.
eg during Daylight Saving Time, 08:30 UTC represents 09:30 BST and therefore the comparison of this datetime to opening hours of 9-5 should be done treating the latter as 08:00 UTC - 16:00 UTC to get accurate results.

The `__contains__` method on `OpeningHours`, that is used by Python for `if dt in object`, continues to use the old behaviour (calling `.available()` with no arguments) for backwards compatibility, so we have switched from that syntax to explicitly calling the `.available()` method with the argument.

This PR also adds a management command to correct records (by simply recalculating the value now) back as far as a given date.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/legalaid/management/commands/recalculate_assigned_out_of_hours.py
+++ b/cla_backend/apps/legalaid/management/commands/recalculate_assigned_out_of_hours.py
@@ -1,0 +1,50 @@
+from django.conf import settings
+from django.core.management import BaseCommand, CommandError
+from django.utils import timezone
+
+from legalaid.models import Case
+
+
+class Command(BaseCommand):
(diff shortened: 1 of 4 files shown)


said-moj (approved):
LGTM 👍 

Non-blocking note:

I think in your PR description you mean BST not BTC:
> during Daylight Saving Time, 08:30 UTC represents 09:30 BTC and therefore the comparison of this datetime to opening

PR #710, opened 2021-06-09
LGA-1721 Delete old cases and complaints

mayowaadeniyi:

- Creates test cases for deleting logs created by users with a **digital.justice.gov.uk** email
- Logs created by users with a non digital.justice.gov.uk email **will not be deleted**
- Let us potentially filter more old cases/complaints to be queued for deletion
- Delete logs by passing a `delete-logs` parameter to the command
- Create conditionals depending on whether the command is called in the terminal/test suite
- Refactor command handling due to linting errors complaining code was too complex

Coveralls failing due to some of the code only accessible/used in the terminal and not in the test suite

## Checklist

### DIFF
--- a/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
+++ b/cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
@@ -3,6 +3,7 @@ from django.core.management.base import BaseCommand
 from dateutil.relativedelta import relativedelta
 
 from legalaid.models import Case
+from cla_eventlog.models import Log
 from cla_butler.tasks import DeleteOldData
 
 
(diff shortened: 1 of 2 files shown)


cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -12,33 +13,57 @@ def get_eligible_cases(self):
...
+            if len(args) > 1 and args[1] == "no-input":
+                self.instance.run()
+            else:
+                answer = self.get_user_input("cases", cases)
+                if answer == "Yes":
+                    self.instance.run()
+        elif args[0] == "delete-logs":
+            digital_justice_user_logs = self.instance.get_digital_justice_user_logs()
+            answer = self.get_user_input("digital justice user logs", digital_justice_user_logs)
+            if answer == "Yes":
```
exonian, line 52: why does this one not have a `no-input` behaviour?

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -33,15 +34,24 @@ def create_case(self, current_time, case_type="legalaid.case"):
         freezer.stop()
         return case
 
-    def create_event_log_for_case(self, case, code, created):
-        make_recipe("cla_eventlog.log", case=case, code=code, created=created)
+    def create_event_log_for_case(self, case, code, created, created_by=None):
+        if created_by:
```
exonian, line 38: I would recommend making a dict for these args
Something like:
```
kwargs = {case:case, code:code, created:created}
if created_by:
    kwargs.append(created_by)
make_recipe("cla_eventlog.log", **kwargs)
```

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
```diff
@@ -228,3 +238,182 @@ def test_old_and_new_cases_are_deleted_correctly(self):
         self.assertEqual(EligibilityCheck.objects.count(), 1)
         self.assertEqual(DiagnosisTraversal.objects.count(), 1)
         self.assertEqual(PersonalDetails.objects.count(), 1)
+
+    def test_log_is_deleted_when_case_is_viewed_by_digital_justice_user(self):
```
exonian, line 242: All of these test cases are extremely verbose and hard to read. Any way of making them shorter by sharing more common stuff and/or just not passing some things (like are the datetimes actually important for these test cases?)

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: Done

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Done. Had to change it to `kwargs.append(created_by=created_by)`

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: I've changed some of the names. Let me know what you think

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
```diff
@@ -12,33 +13,59 @@ def get_eligible_cases(self):
...
+        if args[0] == "delete":
+            if len(args) > 1 and args[1] == "no-input":
+                self.instance.run()
+            else:
+                answer = self.get_user_input("cases", cases)
+                if answer == "Yes":
+                    self.instance.run()
+        elif args[0] == "delete-logs":
+            if len(args) > 1 and args[1] == "no-input":
+                self.instance.run()
```
exonian, line 49: This is wrong. You copy-pasted the condition and didn't change the line that runs

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
exonian: You've only changed the names. I may have been unclear, but it's the test cases that are too long and hard to read. As I say, if you break some of it down into functions, even if you won't have fewer lines of code overall (but you probably will, cos of reusing the function multiple times) it'll mean you can read what the code does without having to read all of how it does it (you can choose to read a function it calls, or not)

exonian (changes requested):
One comment on a mistake.

Re the readability, it wasn't about test names at all. It's just a huge block of lines of code for each test, all repeated for each of the several tests. It needs breaking up so that one can read it (my eyes almost literally refuse!), and really so that you can easily read through a given test case and see what it does, which isn't very possible when the test case is 20-40 lines of unstructured code

cla_backend/apps/cla_eventlog/management/commands/find_and_delete_old_cases.py
mayowaadeniyi: Ty 🤦

cla_backend/apps/cla_eventlog/tests/test_find_and_delete_old_cases.py
mayowaadeniyi: Will respond to the other comment

Pheonnexx (approved):
A lot to read through but it looks good to me!

exonian (approved):

PR #720, opened 2021-06-22
Read SECRET_KEY from environment variable

said-moj:

Read SECRET_KEY from environment variable

Ideally I wanted this to be `os.environ["SECRET_KEY"]` but circleci is not detecting the environment variable that determines which setting file to use. This can be done separately as it probably requires more time to investigate why this is happening and this work is required to go out as soon as possible.

Possible impact after merging this PR:
Users being logged out

## Checklist

### DIFF
--- a/cla_backend/settings/base.py
+++ b/cla_backend/settings/base.py
@@ -170,7 +170,7 @@ STATICFILES_FINDERS = (
 )
 
 # Make this unique, and don't share it with anybody.
-SECRET_KEY = "o=>4$)9?38N@^}d&pj,VL9^{r][xM9L.9cfE:xZZNk(N?v27+i"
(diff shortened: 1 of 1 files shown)


exonian (approved):
Re your point about using `os.environ["SECRET_KEY"]`, I think it's possible that it _is_ using the settings file you intend but that it goes: use specialised settings file; specialised settings file imports everything from base settings file; base settings file looks up env var and throws an exception. If your specialised settings file overrides the value of SECRET_KEY, we never get to it.

The solutions, as far as I can tell, are:
- do similar to what we do here, but default the env var lookup to None, then check at the end of all settings being calculated that we have a value set (which would allow the overriding to work)
- stop overriding the Python vars in our settings, and instead have our overrides being about setting env vars (using something like a .env file). This would mean that the env var would be set on circle and we could enforce with a `os.environ["SECRET_KEY"]`

PR #721, opened 2021-06-29
LGA-1500 Add guidance notes fixture as part of data to be loaded into DB

mayowaadeniyi:

- Adds guidance notes fixture `initial_guidance_notes.json` as part of data to be loaded into DB
- Updated README to reflect this change (will require us to run `./manage loaddata initial_guidance_notes.json` command to apply the change)

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ Create an admin user by running the following command and specifying username ==
 
 Load initial data:
 
-    ./manage.py loaddata initial_groups.json kb_from_knowledgebase.json initial_category.json test_provider.json test_provider_allocations.json initial_mattertype.json test_auth_clients.json initial_media_codes.json test_rotas.json
+    ./manage.py loaddata initial_groups.json kb_from_knowledgebase.json initial_category.json test_provider.json test_provider_allocations.json initial_mattertype.json test_auth_clients.json initial_media_codes.json test_rotas.json initial_guidance_notes.json
 
 Start the server:
 
--- a/app.json
+++ b/app.json
@@ -4,7 +4,7 @@
   "repository": "https://github.com/ministryofjustice/cla_backend",
   "success_url": "/admin/",
   "scripts": {
-    "postdeploy": "./manage.py migrate --noinput && ./manage.py loaddata initial_groups.json kb_from_knowledgebase.json initial_category.json test_provider.json test_provider_allocations.json initial_mattertype.json test_auth_clients.json initial_media_codes.json test_rotas.json"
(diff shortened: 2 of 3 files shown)


exonian (commented):
Looks good.
The clue to what app.json is for is in line 10. I think you're right to update it.


> will require us to rerun the ./manage loaddata -fixtures- command to apply the change)

Could you just clarify exactly what that command is going to look like, please? `-fixtures-` is ambiguous, and I want to be sure we're on the same page

mayowaadeniyi:
> Looks good.
> The clue to what app.json is for is in line 10. I think you're right to update it.
> 
> > will require us to rerun the ./manage loaddata -fixtures- command to apply the change)
> 
> Could you just clarify exactly what that command is going to look like, please? `-fixtures-` is ambiguous, and I want to be sure we're on the same page

Full command will look like this (like how it does in README with my changes):

> ./manage.py loaddata initial_groups.json kb_from_knowledgebase.json initial_category.json test_provider.json test_provider_allocations.json initial_mattertype.json test_auth_clients.json initial_media_codes.json test_rotas.json initial_guidance_notes.json

And ah ok. When we move to Kubernetes fully, will we still need this? I thought all our hosting was done on Kubernetes, plus this exact command is run in `create_db.sh`

exonian:
> Full command will look like below (like how it does in README with my changes):

Do you need to run all of that? Is it desirable to? Does it matter?

exonian:
> And ah ok. When we move to Kubernetes fully, will we still need this? I thought all our hosting was done on Kubernetes, plus this exact command is run in create_db.sh

I have no idea what the "this" in "will we still need this" is

mayowaadeniyi:
> Do you need to run all of that? Is it desirable to? Does it matter?

No I just mentioned it, just so we are aware

> I have no idea what the "this" in "will we still need this" is

So my question was we will still need the `app.json` file? Are we still using heroku as part of our app?

exonian:
> No I just mentioned it just so we know

So what do you propose that we _do_ run?

> So my question was we will still need the app.json file? Are we still using heroku as part of our app?

We don't use heroku at all for template deploy or kubernetes, no.

mayowaadeniyi:
> So what do you propose that we _do_ run?

After having a little think about it, we will just need to run ./manage loaddata initial_guidance_notes.json as this is the only data that is not in the DB (everything else is)
> We don't use heroku at all for template deploy or kubernetes, no.

To be honest deciding whether to delete `app.json` feels out of scope for this ticket anyway, so I think nvm

exonian:
Super. I agree on both counts. If you update the description to reflect both those things, I'll approve it 👍

PR #723, opened 2021-07-02
LGA 1707 Switching on new report

Pheonnexx:

amended env var to be configmap and added that to the helpers so we can change the env var more easily for reports.

Intentionally left blank.

## Checklist


exonian (commented):
Do we want this config to be optional (outside of dev, where it already is) or are we going to always ensure it's present?

Pheonnexx:
I have added optional to the configmap.

exonian (approved):

PR #724, opened 2021-07-06
LGA-1713 - Move frontend hostname and express server port into config

said-moj:

Move frontend hostname and express server port into config

These were stored as secrets before but these values are not secretive values

## Checklist


Pheonnexx (approved):
LGTM 👍 🦖

PR #726, opened 2021-07-07
LGA 1739 remove email addresses from db scripts

Pheonnexx:

Removed email addresses from fixture json file used in create_db

Intentionally left blank.

## Checklist


exonian (approved):
Rubbish PR name but otherwise fine :)

Non-blocking: Is there any benefit to setting them to empty strings rather than omitting the line? Are the model defaults correctly set?

PR #730, opened 2021-07-29
LGA 1459 - Coverage accuracy improvements

Pheonnexx:

I have amended what gets omitted from the coverage report.

It no longer counts tests, migrations, settings and other files that are used but not generally covered by unit tests.

This will unfortunately lower our coverage by roughly 8% however this will more accurately reflect our test coverage.

Intentionally left blank.

## Checklist

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,7 +7,6 @@ omit =
         *tests*
         *apps.py
         *manage.py
-        *__init__.py
         *asgi*
         *wsgi*
         *admin.py
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,3 +3,12 @@ source = cla_backend
 omit =
         *migrations*
         cla_backend/settings/*
+        ./venv/*
+        *tests*
+        *apps.py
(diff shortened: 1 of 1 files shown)


.coveragerc
```diff
@@ -3,3 +3,12 @@ source = cla_backend
 omit =
         *migrations*
         cla_backend/settings/*
+        ./venv/*
+        *tests*
+        *apps.py
+        *manage.py
+        *__init__.py
```
exonian, line 10: This one gives me pause as I suspect we do have real code in some `__init__.py` files that I'd ideally want to see covered.

Is there a particular downside to having these covered, even when they're only the shebang etc?

exonian (commented):
This is really good, although I've commented on the one I'm not 100% on

.coveragerc
Pheonnexx: There isnt, made sure to check and removed from the omit list.

exonian (approved):
Super

PR #732, opened 2021-08-05
Reduce Sentry traces sample rate

thomasleese:
This reduces the Sentry traces sample rate by a half to avoid transactions across MoJ being rate limited. This month [we've hit our rate limit](https://sentry.io/organizations/ministryofjustice/stats/?dataCategory=transactions&pageStatsPeriod=14d&transform=periodic) only half way through the month so I'm looking to reduce the sample rate across our apps which send the most to Sentry. https://github.com/ministryofjustice/hmpps-book-secure-move-api/pull/1529 is an example for `hmpps-book-secure-move-api`.

I hope reducing the sample rate by a half doesn't affect your application too much.

### DIFF
--- a/cla_backend/settings/base.py
+++ b/cla_backend/settings/base.py
@@ -278,7 +278,7 @@ if "SENTRY_DSN" in os.environ:
     sentry_sdk.init(
         dsn=os.environ.get("SENTRY_DSN"),
         integrations=[DjangoIntegration()],
-        traces_sample_rate=1.0,
+        traces_sample_rate=0.5,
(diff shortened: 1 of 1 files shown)


thomasleese:
@said-moj :wave: I think you last edited the code in this repo, would you be able to take a look at this PR?

exonian:
Some stuff's not working on our CI pipeline and I wonder if it's cos the branch isn't ours, so I've created a new feature branch that I'll merge this into and then if that works I'll merge it into our main branch 👍

PR #749, opened 2021-12-23
LGA-1806 - Add disregards in the eligibility report

said-moj:

Add disregards in the eligibility report

If Yes is given as answer to any disregards then use a value of "Yes" or  a value of "No" in the eligibility report

Due to the disregards column being an json object and not an array, I couldn't find a way to check if the object contained any True value within the same main query.

If the column was a json array or a postgres then I could have used a function like the ANY function to check if it contained a True value.

The other option was to introduce a new has_any_disregards column to contain True if any disregards were selected but I think the chosen approach does this without the additional column.

## Checklist


Pheonnexx (approved):
LGTM

PR #769, opened 2022-02-03
Increase timeoutSeconds for worker pods probes

said-moj:

Increase timeoutSeconds for worker pods probes

It seems that on the new cluster the worker pod probes were taking longer than 2 seconds and was causing the pod to crash loop.

## Checklist


Pheonnexx (approved):
LGTM

PR #773, opened 2022-02-15
Some tests misusing assertTrue for comparisons fix

### LINKED ISSUE #772: Some tests misusing assertTrue for comparisons
`assertTrue` is not for comparing arguments, should use `assertEqual` for that.

The developer's intent of the test was to compare argument 1 with argument 2, which is not happening. Really what is happening is the test is passing because first argument is truthy. The correct method to use is assertEqual. [more details](https://codereview.doctor/features/python/best-practice/avoid-misusing-unittest-assert-true)

https://github.com/ministryofjustice/cla_backend/blob/6ce7cb03fbaa68cacd5c6af92c46316b57b57fea/cla_backend/apps/legalaid/tests/test_models.py#L814

I found this issue automatically, see other issues [here](https://codereview.doctor/ministryofjustice/cla_backend)

code-review-doctor:
Fixes #772

### DIFF
--- a/cla_backend/apps/legalaid/tests/test_models.py
+++ b/cla_backend/apps/legalaid/tests/test_models.py
@@ -811,7 +811,7 @@ class CaseTestCase(TestCase):
             case.personal_details.case_count shouldn't get updated
         """
         case = make_recipe("legalaid.case")
-        self.assertTrue(case.personal_details, None)
+        self.assertIsNone(case.personal_details)
 
     def test_case_count_gets_updated_if_pd_not_null(self):
         pd = make_recipe("legalaid.personal_details")


code-review-doctor:
Is the test failing because the test is wrong, or the test logic the test is checking is wrong?

marcofucci:
@code-review-doctor thanks for fixing the typo in the test!

I haven't checked but I think you need to change
`case = make_recipe("legalaid.case")`
into something like
`case = make_recipe("legalaid.case", personal_details=None)`

As the recipe has a default value for personal_details.

The test just makes sure the code doesn't break if personal_details is None and case is saved as there's some custom logic on model save. Hopefully it won't break 😅.

PR #800, opened 2022-05-24
Callback fixture fixes

Pheonnexx:

Fixes for the data to show correctly on case list and have diagnosis and person details.

Intentionally left blank.

## Checklist


jogilder (approved):
Noting addition of case logs but noting that they do not appear in front end interface. Further changes will be necessary but approving now so can get the rest into master as blocking end to end tests.

PR #816, opened 2022-07-04
LGA-2187 - Split test_provider.json into two separate fixtures

said-moj:
Split test_provider.json into two separate fixtures

The fixture test_provider.json contained both provider organisations and provider users belonging to those organisations. 

This PR removes the provider users from the test_provider.json fixture and into a new test_provider_users.json fixture.
The test_provider_users.json and test_auth_clients.json fixtures are only loaded when the `LOAD_TEST_DATA` 
environment variable is to True which is the case on non-public environments.

Seeding data such as provider organisations are loaded when the environment variable `LOAD_SEED_DATA` is set to True.

Test provider organisation need to be available on training but test provider users should not.
This is because training is a public environment.

## Checklist


Pheonnexx (approved):
There is a problem with the fixture upload, keys are clashing so we might need to update them.

jogilder (approved):
LGTM

PR #824, opened 2022-07-12
Feature/lga 1861 cb1 run report fox admin

jogilder:

Updates the code that creates and downloads reports so that they work locally - does not go off and look for AWS
Also updated the version of python specified with black as the higher version (3.9) caused errors with asyncio.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/reports/tests/test_views.py
+++ b/cla_backend/apps/reports/tests/test_views.py
@@ -1,3 +1,4 @@
+import os
 from mock import patch, MagicMock, mock_open
 from django.test import TestCase
 from reports.views import download_file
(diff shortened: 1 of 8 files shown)


cla_backend/apps/reports/views.py
```diff
@@ -184,19 +184,29 @@ def all_knowledgebase_articles(request):
...
+
+        for key, val in headers.items():
+            response[key] = val
+    else:
 
-    k.open_read()
-    headers = dict(k.resp.getheaders())
-    response = HttpResponse(k)
+        def _filepath(filename):
+            return os.path.join(settings.TEMP_DIR, filename)
```
said-moj, line 205: This might look a bit dangerous, as we are returning them any file they want, nothing stopping them traversing backwards to a specific file(i.e ../../../../../)

said-moj (approved):
LGTM 👍

cla_backend/apps/reports/tests/test_views.py
```diff
@@ -0,0 +1,33 @@
...
+        with patch("__builtin__.open", mock_open(read_data="data")) as mock_file:
+            mock_request = MagicMock()
+            # if file_name contains string "schedule"
+            # delete from the database doesn't occur.
+            file_name = "scheduled"
+            settings.AWS_REPORTS_STORAGE_BUCKET_NAME = ""
+            settings.DEBUG = True
+            # download_file(request, file_name="", *args, **kwargs):
+            download_file(mock_request, file_name)
+            assert open("path/to/open").read() == "data"
```
said-moj, line 20: I am not sure which part of reports.views.download_file this is testing?

cla_backend/apps/reports/tests/test_views.py
```diff
@@ -0,0 +1,33 @@
...
+        assert not mock_s3.called
+        mock_file.assert_called_with("path/to/open")
```
said-moj, line 22: Shouldn't this be checking if it was called with the filename that we passed to reports.views.download_file?

cla_backend/apps/reports/tests/test_views.py
psweeting1: Ah yes, this can actually be removed as it proves nothing and is not needed. It doesn't prove anything now I have a much better understanding behind writing unit tests in Python.

So all this is doing is asserting when you call the Open method it reads the mock read_data.... Then the assert after just checks it was called with that path. It has nothing to do with the test... So removed.

Good catch, it's redundant code.

said-moj (approved):
👍 LGTM

PR #836, opened 2022-08-11
LGA-2123 v1.22 Update ingress for Kubernetes 

jogilder:

Creates new ingress file to do upgrade prior to move to Kubernete 1.22.  Helm will assume you want to delete the existing ingress if you only change the current file to create a new ingress.

Helm adds  it’s own annotations: `app.kubernetes.io/managed-by: Helm` and `helm.sh/chart: cla-backend-0.1.3`; because of this if you make changes directly to  <old-ingress-name> and recommit, helm considers the ingress to be a new version of the same resource and deletes the old one. This leads to downtime.

This is a multistep process - first weighting is set to zero for existing 1.21 ingresses, then once this new file has been added and it is verified that traffic is passing through the new v1.22 ingress, the old ingress can be deleted in kubectl and then a further two separate commits will be required to 1) delete the old 1.21 ingress file and then 2)rename this new file as `ingress.yaml`.

## Checklist

### DIFF
--- a/helm_deploy/cla-backend/templates/ingress_v122.yaml
+++ b/helm_deploy/cla-backend/templates/ingress_v122.yaml
@@ -1,0 +1,45 @@
+{{- if .Values.ingress.enabled -}}
+{{- $fullName := include "cla-backend.fullname" . -}}
+{{- $ingressName := printf "%s-%s" $fullName "v122" -}}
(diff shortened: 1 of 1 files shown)


Pheonnexx (approved):
LGTM

said-moj (approved):
👍 LGTM

PR #849, opened 2022-09-08
Feature/lga 2236 Update Opening hours to use env variables

jogilder:

Updated all of the rota opening hours to use environment variables. This means they can be set without changing the code.
There are defaults set in case these are not set.
This change is primarily so we can run out of hours end to end tests.

Have also removed the code creating a  JULY_EDUCATION_FEATURE_FLAG, this will need to be removed from the configMap as well.

## Checklist

### DIFF
--- a/cla_backend/settings/base.py
+++ b/cla_backend/settings/base.py
@@ -5,7 +5,6 @@ import os
 import sentry_sdk
 from boto.s3.connection import NoHostProvided
 from cla_common.call_centre_availability import OpeningHours
-from cla_common.constants import OPERATOR_HOURS
 from cla_common.services import CacheAdapter
 from collections import defaultdict
 from kombu import transport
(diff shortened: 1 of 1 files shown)


cla_backend/settings/base.py
```diff
@@ -52,6 +51,15 @@ def env_var_truthy_intention(name):
 EMAIL_FROM_ADDRESS = "no-reply@civillegaladvice.service.gov.uk"
 DEFAULT_EMAIL_TO = "cla-alerts@digital.justice.gov.uk"
 
+# Want to alter start and end times via environment variables.
+# In case these are not set, default values are set here
+DEFAULT_NON_ROTA_START_TIME_HR = 8
+DEFAULT_NON_ROTA_END_TIME_HR = 17
```
psweeting1, line 57: DEFAULT_NON_ROTA_END_TIME_HR and DEFAULT_ED_END_TIME_HR are the same values. Do we need two seprate variables? 

Same could be said for DEFAULT_DISCRIM_START_TIME_HR and DEFAULT_NON_ROTA_START_TIME_HR. 

Just feels like we're repeating ourselves.

jogilder: They are currently set to the same value but they could be different as they are used for different categories of provider. Hence we need the option of different variables to allow overrides for the various categories.

cla_backend/settings/base.py
```diff
@@ -52,6 +51,15 @@ def env_var_truthy_intention(name):
...
+DEFAULT_ED_START_TIME_HR = 9
+DEFAULT_ED_END_TIME_HR = 17
+DEFAULT_DISCRIM_START_TIME_HR = 8
+DEFAULT_DISCRIM_END_TIME_HR = 18
```
said-moj, line 61: Can we move these closer to where they are used?

psweeting1 (approved):
LGTM, would be good to get Heather approval as well as she has better knowledge than me.

said-moj (approved):
👍 LGTM

PR #850, opened 2022-09-09
LGA 2222 Replace django oauth2 provider

Pheonnexx:

This is replacing the package django oauth2 provider with the django oauth2 toolkit.

We have had to tweak and transfer over some methods, so I have taken the opportunity to split them up and help clarify what they do.

The unit testing was actually pretty good on this section so has remained the same.

I have cleared out a form which is no longer used with the toolkit.

Toolkit is currently at 0.11.0 which is the max version it could be for the version of django that we have.

Intentionally left blank.

## Checklist


cla_backend/apps/knowledgebase/fixtures/kb_from_spreadsheet.json
```diff
@@ -0,0 +1,185 @@
+[
```
said-moj, line 1: Do we need this file?

requirements/generated/requirements-base.txt
```diff
@@ -0,0 +1,189 @@
+#
```
said-moj, line 1: Was this generated manually by accident as we don't use requirements/generated/requirements-base.txt?

cla_backend/apps/cla_auth/views.py
```diff
@@ -16,6 +26,10 @@
 class AccessTokenView(Oauth2AccessTokenView):
     throttle_classes = [LoginRateThrottle]
 
+    def __init__(self, *args, **kwargs):
+        super(AccessTokenView, self).__init__(*args, **kwargs)
+        self.account_lockedout = False
```
said-moj, line 31: What is the purpose of this `self.account_lockedout` variable?
From what I can see it is set in [check_login_attempts](https://github.com/ministryofjustice/cla_backend/pull/850/files#diff-d415eef3965a9884491996b654d9f2a87f0dd5c732c8658d7c17b5b4a8de71e0R156) but that throws an exception and stops execution and doesn't continuing doing any further checks when attempts are more than settings.LOGIN_FAILURE_LIMIT

said-moj:
@Pheonnexx does this include the migration of the existing provider.clients into new application model?

mtac50:
There are a few drops in test coverage, is there anything we can do to rectify it?

requirements/generated/requirements-base.txt
Pheonnexx: removed

mtac50 (changes requested):

bin/create_db.sh
```diff
@@ -48,12 +48,19 @@ admin_password() {
     fi
 }
 
+copy_old_client_data_to_new_table() {
+    if [ -n "$CLIENT_DATA_COPY" ]; then
```
mtac50, line 52: I'm being blind, this should be ` if [ "$CLIENT_DATA_COPY" == "True" ]; then`

helm_deploy/cla-backend/values-uat.yaml
```diff
@@ -55,3 +55,5 @@ envVars:
     value: "cla-frontend-socket-server.laa-cla-frontend-uat.svc.cluster.local"
   FRONTEND_HOST_NAME:
     value: "http://cla-frontend-socket-server.laa-cla-frontend-uat.svc.cluster.local"
+  MIGRATE_OAUTH_DATA:
+    value: "False"
```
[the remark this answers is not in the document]
mtac50: While we use a local Postgres db for the UAT deployments it doesn't need to be `True` as the oauth data added via fixtures when the db is created.

cla_backend/apps/cla_auth/management/commands/copy_client_data_to_new_table.py
```diff
@@ -0,0 +1,73 @@
...
+    )
+
+    def handle(self, *args, **options):
+        clients = self.get_queryset()
+        try:
+            for client in clients:
+                user = None
+
+                if client["user_id"] is not None:
+                    user = User.objects.get(pk=client["user_id"])
```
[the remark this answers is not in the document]
mtac50: The `user` field int he `Application` model doesn't have a `NOT NULL` constraint so if `client["user_id"]` was `None` then then the default `user=None` is used which _shouldn't_ cause any errors (famous last words...).

cla_backend/apps/cla_auth/views.py
```diff
@@ -40,31 +54,111 @@ def throttled(self, request, wait):
...
+        return class_name
+
+    def check_login_attempts(self, request):
+        """
+        This checks how many login attempts there have been, it locks the user out when the
+        LOGIN_FAILURE_LIMIT has been hit.
+        """
+        username = request.POST.get("username")
+
+        cooloff_time = timezone.now() - datetime.timedelta(minutes=settings.LOGIN_FAILURE_COOLOFF_TIME)
```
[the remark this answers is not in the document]
mtac50: Good spot. Done.

psweeting1 (approved):
LGTM

cla_backend/apps/cla_auth/tests/management/commands/test_copy_client_data_to_new_table.py
```diff
@@ -0,0 +1,85 @@
...
+        Application.objects.create(
+            user=self.test_user,
+            name=self.OLD_CLIENT_DATA["name"],
+            client_type=self.OLD_CLIENT_DATA["client_type"],
+            client_id=self.OLD_CLIENT_DATA["client_id"],
+            client_secret=self.OLD_CLIENT_DATA["client_secret"],
+            redirect_uris=self.OLD_CLIENT_DATA["redirect_uri"],
+            authorization_grant_type="password",
+        )
+        with atomic():
```
jogilder, line 71: why does this use atomic and the other check does not?

mtac50: The run of the management command in this test intentionally triggers an `IntegrityError` which was causing issues as  django test cases are all wrapped up in their own transaction. The use of atomic here makes the command run as a nested transaction to avoid errors.

PR #857, opened 2022-10-06
LGA-2164 Roll back database-11 changes

jogilder:

When comparing pre and post migration reports, it seems as though not all the files were not the same. This PR makes the application point back at database-10 so can see if there was an issue with the migration.

Trying to rule out the issue being with this or the oauth change.

Intentionally left blank.

## Checklist


psweeting1 (approved):
LGTM

PR #860, opened 2022-10-10
LGA 2213 Delete old/unneeded cases

Pheonnexx:

amended task so instead of deleting tasks over two years, it waits till they are three years old and do not have a specific outcome_code.
 
I amended/added unit tests for the task.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/cla_butler/tasks.py
+++ b/cla_backend/apps/cla_butler/tasks.py
@@ -50,6 +50,7 @@ class DeleteOldData(Task):
     We also delete empty cases and data thet is not connected to anything in
     particular.
     """
+    OUTCOME_CODES = ["COI", "MIS", "REOPEN", "SPOP"]
 
     def __init__(self, *args, **kwargs):
         self._setup()
@@ -74,9 +75,8 @@ class DeleteOldData(Task):
         outcome code indicating its closed.
         """
         three_years = self.now - relativedelta(years=3)
-        # outcome_codes list is temporary and a work in progress
-        outcome_codes = ["COI", "MIS", "REOPEN", "SPOP"]
-        return Case.objects.filter(modified__lte=three_years).exclude(outcome_code__in=outcome_codes)
+        
+        return Case.objects.filter(modified__lte=three_years).exclude(outcome_code__in=self.OUTCOME_CODES)
 
     def _delete_logs(self, qs):
         ct = ContentType.objects.get_for_model(qs.model)
--- a/cla_backend/apps/cla_butler/tasks.py
+++ b/cla_backend/apps/cla_butler/tasks.py
@@ -47,21 +47,8 @@ class DeleteOldData(Task):
     """
     Deletes old data that is no longer needed.
 
-    Case data more than 2 years old is no longer needed so will be deleted.
-
     We also delete empty cases and data thet is not connected to anything in
     particular.
-
-    Maybe faster to dump to json using
-    from django.core.serializers.json import DjangoJSONEncoder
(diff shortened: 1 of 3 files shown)


cla_backend/apps/cla_butler/tests/test_tasks.py
```diff
@@ -127,7 +130,41 @@ def test_delete_old_data_run_case_over_two_years_successful_delete(self):
...
+        """
+        log = make_recipe("cla_auditlog.audit_log")
+
+        # Creating a case thats three years old so it gets picked up properly by the delete data
+        freezer = freeze_time(timezone.now() + relativedelta(years=-4))
+        freezer.start()
+        case = make_recipe("legalaid.case", audit_log=[log], outcome_code="SPOP")
+        freezer.stop()
+
+        eod = make_recipe("legalaid.eod_details", case=case)
```
mtac50, line 146: The assertion code for all three test methods is the same, can we refactor it into a common method where we pass in the expected number of object before/after to avoid duplication?

cla_backend/apps/cla_butler/tasks.py
```diff
@@ -81,6 +68,16 @@ def run(self, *args, **kwargs):
...
         self.now = timezone.now()
 
+    def get_eligible_cases(self):
+        """
+        This gets cases which are over three years old and have a specific
+        outcome code indicating its closed.
+        """
+        three_years = self.now - relativedelta(years=3)
+        # outcome_codes list is temporary and a work in progress
+        outcome_codes = ["COI", "MIS", "REOPEN", "SPOP"]
```
mtac50, line 78: Can this be a class constant so we can access the values in the tests?

cla_backend/apps/cla_butler/tests/test_tasks.py
```diff
@@ -127,7 +130,41 @@ def test_delete_old_data_run_case_over_two_years_successful_delete(self):
...
+        """
+        This makes sure that even if a case is over 3 years
+        its not deleted due to having an excluded outcome_code
+        """
+        log = make_recipe("cla_auditlog.audit_log")
+
+        # Creating a case thats three years old so it gets picked up properly by the delete data
+        freezer = freeze_time(timezone.now() + relativedelta(years=-4))
+        freezer.start()
+        case = make_recipe("legalaid.case", audit_log=[log], outcome_code="SPOP")
```
mtac50, line 143: For completeness I would create a case for each outcome code on the exclusion list.

cla_backend/apps/cla_butler/tests/test_tasks.py
```diff
@@ -127,7 +130,41 @@ def test_delete_old_data_run_case_over_two_years_successful_delete(self):
...
+        self.delete_old_data.run()
+
+        self.assertEqual(len(Case.objects.all()), 1)
+        self.assertEqual(len(AuditLog.objects.filter(case__in=pks)), 1)
+        self.assertEqual(len(EODDetails.objects.all()), 1)
+        self.assertEqual(len(Complaint.objects.all()), 1)
+        case_complaints = Complaint.objects.filter(eod_id__in=eods).values_list("pk", flat=True)
+        self.assertEqual(len(AuditLog.objects.filter(complaint__in=case_complaints)), 1)
+
+    def test_delete_old_data_run_case_under_three_years_unsuccessful_delete(self):
```
mtac50, line 167: In this test, is it worth giving the case an outcome code that we no should be delete if it was over 3 years old?

mtac50 (approved):
Looks good!

PR #862, opened 2022-10-19
LGA-2291: Update migration validation script

mtac50:

- [x] `pgDataDiff` installation added to the database migration dockerfile.
- [x]  Migration validation script updated to compare the database contents using the `pgDataDiff` tool.

- `pgDataDiff` requests `md5` hashes for "chunks" of data and using this to do a fast comparison. At the moment the `chunk_size` has been set to 100k. This means that the comparison is fast but we won't be able to pinpoint the exact row where the difference exists.

### DIFF
--- a/bin/database-migration/migration_validation.py
+++ b/bin/database-migration/migration_validation.py
@@ -3,6 +3,8 @@ import sys
 import os
 import psycopg2
 
+from pgdatadiff.pgdatadiff import DBDiff
+
 
 def eprint(*args, **kwargs):
     print(*args, file=sys.stderr, **kwargs)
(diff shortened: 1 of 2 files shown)


bin/database-migration/migration_validation.py
```diff
@@ -91,6 +93,29 @@ def get_last_row(cursor, table_name):
     return cursor.fetchone()
 
 
+def check_table_data(error_list):
```
said-moj, line 96: Does this need the error_list if it's returning it?
Maybe have it just return the error_list and then append it to your errors after you call it

said-moj (approved):
👍 LGTM

PR #875, opened 2022-11-09
LGA-2122: Set session timeout for fox admin.

mtac50:

- [x] Added a session timeout of 30mins to fox admin 
- [x] Session timeout is request with every request so the timeout only applies during inactivity.
- [x] Set the fox admin session to reset when the browser is closed.

## Additional Notes:

There was concern from users about how this would effect report generation which is sometimes longer than 30mins. I have checked the code and it looks like the report generation is run asynchronously by Celery so the session timeout will not effect the generation itself however the users will stilled be logged out if idle for 30mins. The reports will appear in the list to download when users log back in.

### DIFF
--- a/cla_backend/apps/cla_auth/tests/test_views.py
+++ b/cla_backend/apps/cla_auth/tests/test_views.py
@@ -1,12 +1,15 @@
 import datetime
 
 from django.contrib.auth.models import User
+from django.contrib.sessions.models import Session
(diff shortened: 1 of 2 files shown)


Pheonnexx (approved):
LGTM 👍  nice test.

said-moj (approved):
👍 LGTM

PR #887, opened 2022-11-18
feature/LGA-2112-FoxAdmin session security added.

mtac50:

- [x] Django-session-security add for the FoxAdmin portal.

- [x] Selenium test cases added for the new functionality.
- [x] firefox added to the circle CI container to run the Selenium tests


psweeting1:
Nice job, looks good. My only feedback is that if this is an integration test, then then can the directory that it's in say it's an integration test e.g. 

`cla_backend/apps/core/tests/integration/admin/test_session_security.py`

cla_backend/apps/core/tests/admin/test_session_security.py
```diff
@@ -0,0 +1,202 @@
...
+        action.perform()
+
+    def confirm_logged_in(self):
+        """Confirms the broswers is logged into the admin portal by check for the logout button
+        and the session cookie.
+        """
+        try:
+            WebDriverWait(self.browser, self.WAIT).until(
+                expected_conditions.presence_of_element_located((By.LINK_TEXT, "Log out")))
+        except Exception:
```
psweeting1, line 154: Do you know what type of exception is thrown? This is me coming from a Java side here, but I think it would be better to catch the correct Exception, instead of a blanket of any Exception, better to be explicit. 

Same goes for all try catches that use Exception

mtac50: Any yes, good shout! Now done :)

psweeting1 (approved):
LGTM

said-moj:
@mtac50 is there a reason why we would have browser based testing in this repo and not in the end-to-end repo?

Pheonnexx (approved):
LGTM 👍

PR #908, opened 2023-02-16
LGA-1776 update django restframework to 3.3.3

MishNajam:

- Pagination has changed for 3.3.3. Rather than use properties of the serializer viewset, these now have to be declared via a pagination class. You cannot use the base pagination class, you have to override and declare your own. The base pagination class has key properties set as none. This means that even if you declared pagination in the viewset by declaring `pagination_class=PageNumberPagination` there would be no pagination.  

- ExportListView had a page_size of 1000 set but this did not seem to be doing anything - when we added pagination in for that viewset, the view did not work as expected. Hence `page_size` was removed and no pagination included

- A manual test to check that more than 1000 rows did not change behaviour between master and this branch was performed.

- ArticleViewSet in checker had settings for pagination that mirrored those in BaseArticleViewSet so these were removed as not adding anything above the base class.

- Fixed language versions in the pre-commit.yaml for all of the hooks - this helps when developers may have different versions of pre-commit installed.
- We used new database diff tool to check for any changes not picked up by the tests and this looked fine. Have agreed to create an annotation/md file for the end to end tests repository to explain why there will currently be some differences. This annotation can then be used to update the diff tool so that it can automatically accept these differences - a JIRA card will be created.

## Checklist

### DIFF
--- a/cla_backend/apps/checker/views.py
+++ b/cla_backend/apps/checker/views.py
@@ -6,7 +6,6 @@ from rest_framework.response import Response as DRFResponse
 
 from core.models import get_web_user
 from core.drf.mixins import ClaCreateModelMixin, ClaUpdateModelMixin
-from core.drf.paginator import StandardResultsSetPagination
 from core.drf.viewsets import CompatGenericViewSet
 from checker.helpers import notify_callback_created
 from diagnosis.views import DiagnosisModelMixin
@@ -45,7 +44,6 @@ class ArticleCategoryNameFilter(ArticleCategoryFilter):
 
 
 class ArticleViewSet(PublicAPIViewSetMixin, BaseArticleViewSet):
-    pagination_class = StandardResultsSetPagination
 
     filter_class = ArticleCategoryNameFilter
 
--- a/cla_backend/apps/checker/views.py
+++ b/cla_backend/apps/checker/views.py
@@ -6,6 +6,7 @@ from rest_framework.response import Response as DRFResponse
 
 from core.models import get_web_user
 from core.drf.mixins import ClaCreateModelMixin, ClaUpdateModelMixin
+from core.drf.paginator import StandardResultsSetPagination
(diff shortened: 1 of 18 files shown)


cla_backend/apps/reports/api.py
```diff
@@ -14,7 +13,6 @@ class ExportListView(generics.ListAPIView, generics.DestroyAPIView):
     model = Export
     permission_classes = (IsAdminUser,)
     authentication_classes = (SessionAuthentication,)
-    page_size = 1000
```
said-moj, line 17: Why wasn't `pagination_class` added to this viewset?

cla_backend/apps/checker/views.py
```diff
@@ -44,8 +44,6 @@ class Meta(ArticleCategoryFilter.Meta):
 
 
 class ArticleViewSet(PublicAPIViewSetMixin, BaseArticleViewSet):
-    paginate_by_param = "page_size"
```
said-moj, line 47: Why wasn't `pagination_class` added to this viewset?

cla_backend/apps/reports/api.py
jogilder: We don't think that this one was being used in master - we added in pagination and then took it out again as adding in pagination actually resulted in a difference to master. We did a one off automated test adding in over 1000 rows to the table of exports and neither master or the new branch seemed to do any pagination.

cla_backend/apps/checker/views.py
jogilder: Not sure, I will follow up on this one

jogilder: Checked and this was not doing anything - BaseArticleViewSet already declares the same pagination details

Pheonnexx (approved):
awesome work 🚀

said-moj (approved):
👍 LGTM

PR #914, opened 2023-03-09
LGA-1777 django upgrade 1 8 19

psweeting1:
Upgrade Django to 1.8.19
https://docs.djangoproject.com/en/1.8/releases/1.8/

Major changes seen in this version of Django:
- Changes to the way templates are managed, have updated the settings as described in the announcement.
- Changes to the permissions for the usertools block - see miscellaneous changes in the release notes: https://docs.djangoproject.com/en/4.1/releases/1.8/#miscellaneous
- Changes to management commands, now use argparse instead of optparse as standard so needed to change the way parameters are passed in for custom commands.
- Changes to TestCase data setup. Removed setUpClass from test_case_api and replaced with setUpTestData that calls cls, resolving the cls atomics issue.  All test setup is now included in `setUpTestData` instead of `setUpClass`. Link to stackoverflow - https://stackoverflow.com/a/29655301
- `select_related` now checks all inputs to see if they are valid. Logs was not a valid input for ComplaintManager so was removed after checking versus master and seeing no change after it's removal.

When running the database diff script from the end to end repository we compared these results with those after the previous dsrf ticket. We found changes to three tables:

- diagnosis_diagnosistraversal - same information across both tables but populated on different rows

- django_content_type - The name column has been removed as part of the Django upgrade. “The name field of django.contrib.contenttypes.models.ContentType has been removed by a migration and replaced by a property. That means it’s not possible to query or filter a ContentType by this field any longer.” - [Django](https://docs.djangoproject.com/en/4.1/releases/1.8/#miscellaneous) 

- Django-migrations table - 6 new migrations. All of these were linked to changes listed in the migrations

    - 2 were related to the removal of the name field in content type.

    - 1 was related to the default email field max_length being increased to 254 characters

    - 1 was related to the max name for Permission.name being increased to 255 characters

    - 1 was related to allowing the user’s last login field to be nullable

    - 1 was related to changing username “opts” - which there is some debate which change in the release notes it relates to

## Checklist

### DIFF
--- a/cla_backend/apps/reports/tests/test_reports.py
+++ b/cla_backend/apps/reports/tests/test_reports.py
@@ -277,7 +277,6 @@ class OBIEEExportOutputsZipTestCase(TestCase):
 
 
 class TestKnowledgeBaseArticlesExport(TestCase):
-    @classmethod
     def setUp(self):
         article_1, article_2, article_3, article4 = make_recipe("knowledgebase.article", _quantity=4)
         make_recipe("knowledgebase.telephone_number", article=article_1, number=123)
--- a/cla_backend/apps/reports/tests/test_reports.py
+++ b/cla_backend/apps/reports/tests/test_reports.py
@@ -278,8 +278,7 @@ class OBIEEExportOutputsZipTestCase(TestCase):
 
 class TestKnowledgeBaseArticlesExport(TestCase):
     @classmethod
-    def setUpClass(cls):
-        super(TestKnowledgeBaseArticlesExport, cls).setUpClass()
+    def setUp(self):
         article_1, article_2, article_3, article4 = make_recipe("knowledgebase.article", _quantity=4)
         make_recipe("knowledgebase.telephone_number", article=article_1, number=123)
         make_recipe("knowledgebase.telephone_number", article=article_1, number=456, name="special")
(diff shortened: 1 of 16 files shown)


cla_backend/apps/reports/tests/test_reports.py
```diff
@@ -261,8 +261,7 @@ def tearDown(self):
 
 
 class TestKnowledgeBaseArticlesExport(TestCase):
-    @classmethod
-    def setUpClass(cls):
+    def setUp(self):
```
MishNajam, line 264: "setUp runs before each test in a test case at the instance level. setUpTestData runs once before each test case at the class level. So if you have a TestCase with N tests, setUpTestData will be called only once (before *all* tests) and setUp will be called N times (before *each* test)."

So I suppose it is down to how we want this test to run. 


https://docs.djangoproject.com/en/4.1/topics/testing/tools/#django.test.TestCase.setUpTestData
https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp

MishNajam: I've changed it to setUpTestData() and kept the @classmethod decorator in there. This is to mimic the prior setUpClass() functionality. This is also in line with the change we made in cla_backend/apps/call_centre/tests/api/test_case_api.py

Pheonnexx (dismissed):
LGTM

said-moj (dismissed):

MishNajam (approved):
LGTM

Pheonnexx (approved):

PR #918, opened 2023-03-16
Adding script to switch backend, frontend and public in and out of maintenance mode

said-moj:

Adding script to switch backend, frontend and public in and out of maintenance mode

Intentionally left blank.

## Checklist


jogilder:
@said-moj as discussed this should allow for when there is no staging etc

PR #926, opened 2023-04-03
LGA-2416 Make provider name unique to prevent CLA provider report fail

MishNajam:

The MI Allocation Provider report fails when there is a duplicate case sensitive provider name. (IE - "Test" and "Test" fails, whereas "Test" and "test" would not fail)

To ensure that provider name is unique (in a case sensitive manner as above), this PR:
- updates the provider model 
- adds a migration file
- adds a unit test 

Updating the model pushes an error to the django form as seen in the screenshot below:
Screenshot 2023-04-03 at 11 36 42 (2)

## Checklist

### DIFF
--- a/cla_backend/apps/cla_provider/migrations/0003_auto_20230405_1449.py
+++ b/cla_backend/apps/cla_provider/migrations/0003_auto_20230405_1449.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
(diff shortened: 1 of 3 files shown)


said-moj (approved):
👍 LGTM

jogilder (commented):
Does this need a migration as we have updated the model?
https://docs.djangoproject.com/en/4.1/ref/django-admin/#django-admin-makemigrations

jogilder (approved):
LGTM

sarahyjja (approved):
👍

PR #941, opened 2023-05-11
LGA 2452 Use Gov Notify for emails

jogilder:

- Update any email sending in cla_backend to use gov.notify as no longer using sendgrid

- Renamed settings.SITE_HOSTNAME to settings.FRONTEND_HOST_NAME as it was incorrectly being used when creating external links inside emails

- Replace usage of django.core.mail.mailbox with govnotfy test mailbox

Emails for operator actions (create or modified) and staff actions (create or modified) have been replaced with govuknotify however they do not actually send emails because the OPERATOR_USER_ALERT_EMAILS in base.py settings is an empty list

## Checklist


cla_backend/apps/govuk_notify/api.py
```diff
@@ -0,0 +1,31 @@
...
+        try:
+            response = self.notifications_client.send_email_notification(
+                email_address=email_address,  # required string
+                template_id=template_id,  # required UUID string
+                personalisation=personalisation,
+            )
+        except HTTPError as error:
+            logger.error("GovUkNotify error: {msg}".format(msg=str(error)))
+            raise error
+        return response
```
MishNajam, line 31: if we never use the response, do we need it?

sarahyjja (approved):
LGTM

PR #957, opened 2023-07-12
LGA-2467 make cla_backend pipelines use postgres v14 image

MishNajam:

As per the investigation completed in the spike LGA-2408, this PR uses the postgres v14 image by:

- updating the circleci postgres image from 11.6 alpine to 14.7 bullseye
- Updating the MISurveyExtract.sql file to pass our unit tests when upgrading to allow upgrade past v11.14 (see highlights for more info)

1. When upgrading from v11.13 to v11.14 2 unit tests trigger and fail:
```
ProgrammingError: in an aggregate with DISTINCT, ORDER BY expressions must appear in argument list
LINE 18: ..._agg(DISTINCT cc_org.name::varchar, ', ' ORDER BY cc_org.nam...
```
Based on the v11.14 documentation this is due to postgres no longer ignoring unspecified cast type modifier:

> Don't discard a cast to the same type with unspecified type modifier (Tom Lane)
> 
> For example, if column f1 is of type numeric(18,3), the parser used to simply discard a cast like f1::numeric, on the grounds that it would have no run-time effect. That's true, but the exposed type of the expression should still be considered to be plain numeric, not numeric(18,3). This is important for correctly resolving the type of larger constructs, such as recursive UNIONs.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,6 +126,7 @@ jobs:
       - image: postgres:14.7-bullseye
         environment:
           POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
           POSTGRES_DB: circle_test
     steps:
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,11 +121,11 @@ jobs:
       - image: cimg/python:2.7
         environment:
           DB_NAME: circle_test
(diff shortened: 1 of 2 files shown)


TawneeOwl (approved):
LGTM

jogilder (approved):
LGTM

PR #961, opened 2023-07-17
LGA-2569 - Force ValidationError message to a string

said-moj:

Force ValidationError message to a string

The previous field implementations did not forcibly coerce returned values into the correct type in many cases. For example, an ValueError would return a string output if used in a string context. We now more strictly coerce to the correct return type, leading to more constrained and expected behavior

https://www.django-rest-framework.org/community/3.0-announcement/#coercing-output-types

## Checklist


jogilder (approved):
LGTM but think we should check if there are other similar ValidationError issues.

psweeting1 (approved):
LGTM, can you also make a e2e test for this as well so we've covered ourselves for future updates.

PR #981, opened 2023-08-24
Split SLA1 and SLA2 into individual CB1 reports

Pheonnexx:

Takes the SLA2 out of the existing CB1 report and into its own CB1 report, as SLA1 and SLA2 are with cases that have different codes rather than just 'call_started'

Intentionally left blank.


psweeting1 (approved):
LGTM

TawneeOwl (approved):
LGTM 🚀

PR #988, opened 2023-09-08
LGA-2673 - Temporarily disable pushing notifications to the socket server

said-moj:

Temporarily disable pushing notifications to the socket server

At the moment after receiving the notification, the socket server does not send the notification to the user due to it being setup incorrectly See https://github.com/ministryofjustice/cla_frontend/pull/919 for more details

Also the scheduling of these notifications with long ETA's is causing duplicating messages(thousands) to fill up the message queue, this means reports or anything else that depends on the message queue will not work until the queue is purged manually

## Checklist

### DIFF
--- a/cla_backend/apps/notifications/models.py
+++ b/cla_backend/apps/notifications/models.py
@@ -2,7 +2,7 @@
 from django.db import models
 from django.conf import settings
 from django.utils import timezone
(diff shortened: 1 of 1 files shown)


psweeting1 (approved):
LGTM Delete everything coming in.

PR #993, opened 2023-09-12
LGA-2673 - Move notifications to to a periodic task

said-moj:

Move notifications to to a [period task](https://docs.celeryq.dev/en/stable/userguide/periodic-tasks.html)

### Periodic task
In our current version of celery(4.0.2), celery collects all the periodic tasks at startup and schedules them for execution at their defined period(i.e every 5 minutes or a specific date). However we need to schedule notifications to be sent out when they are due, with the celery limitation this means we cannot schedule notifications as their own periodic tasks.
Instead we have one periodic task that runs every 3 minutes and sends out any notifications that have their due date in the past but have not been sent yet.

#### Why are you not using [django-celery-beat](https://github.com/celery/django-celery-beat)
We tried using django-celery-beat but due to our django/celery version we had to us an older version(1.3) and it was not working as expected. It was not scheduling created/updated periodic tasks after startup.

### Why change from task to Periodic task
Previously we were scheduling notification as celery tasks with a very long ETA. However this was causing duplicate notifications tasks to build up in the message queue to the point where there were thousands of duplicate notification tasks which meant other tasks like reports could not run until the queue was cleared of the duplicate notifications.

From the celery documentation tasks with long ETAs are discouraged.

> Warning
Tasks with eta or countdown are immediately fetched by the worker and until the scheduled time passes, they reside in the worker’s memory. When using those options to schedule lots of tasks for a distant future, those tasks may accumulate in the worker and make a significant impact on the RAM usage.
> 
> Moreover, tasks are not acknowledged until the worker starts executing them. If using Redis as a broker, task will get redelivered when countdown exceeds visibility_timeout (see [Caveats](https://docs.celeryq.dev/en/stable/getting-started/backends-and-brokers/redis.html#redis-caveats)).
> 
> Therefore, using eta and countdown is not recommended for scheduling tasks for a distant future. Ideally, use values no longer than several minutes. For longer durations, consider using database-backed periodic tasks, e.g. with [django-celery-beat](https://pypi.python.org/pypi/django-celery-beat/) if using Django (see [Using custom scheduler classes](https://docs.celeryq.dev/en/stable/userguide/periodic-tasks.html#beat-custom-schedulers)).

https://docs.celeryq.dev/en/stable/userguide/calling.html#eta-and-countdown

## Checklist


cla_backend/apps/notifications/periodic_tasks.py
```diff
@@ -0,0 +1,59 @@
...
+    data = json.dumps(data, cls=DjangoJSONEncoder)
+    logger.info("Notifications: url %s" % url)
+    logger.info("Notifications: data %s" % data)
+    response = requests.post(
+        url,
+        data=data,
+        headers={"content-type": "application/json"},
+    )
+    success = response.json()["success"]
+    assert success is True
```
psweeting1, line 58: Would it not be a status to check here?

`response.status_code == 200` Something like that.

said-moj: The socket server returns `{...,"success": true}` when it has completed successfully

psweeting1 (approved):
LGTM

BenMillar-CGI (approved):
Looks good to me 👍

PR #994, opened 2023-09-14
LGA 2664 Extend MICaseExtract with notes and provider notes

Pheonnexx:

Copied MICaseExtract and made MICaseExtractExtend, this adds on notes and provider notes, as this could affect the length of the report running time and anyone who uses the original report.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/reports/forms.py
+++ b/cla_backend/apps/reports/forms.py
@@ -296,6 +296,106 @@ class MICaseExtract(SQLFileDateRangeReport):
         sql = self.query.format(diversity_expression=diversity_expression)
         sql_args = [passphrase] + list(self.date_range)
         return self.execute_query(sql, sql_args)
+    
+    
+class MICaseExtractExtended(SQLFileDateRangeReport):
+    QUERY_FILE = "MIExtractByOutcomeExtended.sql"
+
+    passphrase = forms.CharField(
+        required=False, help_text="Optional. If not provided, the report will not include diversity data"
+    )
+
+    def get_headers(self):
+        return [
+            "LAA_Reference",
+            "Hash_ID",
+            "Case_ID",
+            "Split_Check",
+            "Split_Link_Case",
+            "Provider_ID",
+            "Category_Name",
+            "Date_Case_Created",
+            "Last_Modified_Date",
+            "Outcome_Code_Child",
+            "Billable_Time",
+            "Cumulative_Time",
+            "Matter_Type_1",
+            "Matter_Type_2",
+            "User_ID",
+            "Scope_Status",
+            "Eligibility_Status",
+            "Adjustments_BSL",
+            "Adjustments_LLI",
+            "Adjustments_MIN",
+            "Adjustments_TYP",
+            "Adjustments_CallbackPreferred",
+            "Adjustments_Skype",
+            "Gender",
+            "Ethnicity",
+            "Age(Range)",
+            "Religion",
+            "Sexual_Orientation",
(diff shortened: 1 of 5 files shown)


cla_backend/apps/reports/urls.py
```diff
@@ -16,6 +16,7 @@
         name="mi_provider_allocation_extract",
     ),
     url(r"^mi-case-extract/$", views.mi_case_extract, name="mi_case_extract"),
+    url(r"^mi-case-extract/$", views.mi_case_extract_extended, name="mi_case_extract_extended"),
```
BenMillar-CGI, line 19: I believe this url should point to mi-case-extract-extended. 

Currently, it is opening the mi-case-extract page.

cla_backend/apps/reports/forms.py
```diff
@@ -298,6 +298,106 @@ def get_queryset(self):
...
+    def get_rows(self):
+        for row in self.get_queryset():
+            full_row = list(row)
+            diversity_json = full_row.pop() or {}
+
+            def insert_value(key, val):
+                index = self.get_headers().index(key)
+                full_row.insert(index, val)
+
+            insert_value("Gender", diversity_json.get("gender"))
```
said-moj, line 379: @Pheonnexx Diversity fields can be return in the sql
https://github.com/ministryofjustice/cla_backend/blob/master/cla_backend/apps/reports/sql/obiee/export_legal_aid_personaldetails.sql#L22

cla_backend/apps/reports/sql/MIExtractByOutcomeExtended.sql
```diff
@@ -0,0 +1,173 @@
...
+  LEFT OUTER JOIN operator_first_access on operator_first_access.case_id = c.id and operator_first_access.rn = 1
+  LEFT OUTER JOIN provider_first_view on provider_first_view.case_id = c.id and provider_first_view.rn = 1
+  LEFT OUTER JOIN provider_first_assign on provider_first_assign.case_id = c.id and provider_first_assign.rn = 1
+  LEFT OUTER JOIN legalaid_case split_case on c.from_case_id = split_case.id
+  LEFT OUTER JOIN cla_provider_provider as assigned_provider on (log.context->>'provider_id')::numeric = assigned_provider.id
+  LEFT OUTER JOIN call_centre_organisation AS cc_org ON cc_org.id = c.organisation_id
+where
+  log.type = 'outcome'
+  and log.created >= %s
+  and log.created < %s
```
BenMillar-CGI, line 173: Just a quick query regarding this line, should this line not read `and log.created <= %s`

With the current implementation, entering the values:
Date from: 15/09/2023
Date to: 15/09/2023

Would always return no results, where the expected behaviour may be to generate a report of all today's cases.

Many thanks

Pheonnexx: This is the current behaviour of the other report that this comes from so reluctant to change this.

BenMillar-CGI: Sounds good, makes sense to match the current report implementation.

BenMillar-CGI (approved):
Looks good to me, notes are added to the CSV as expected in UAT 👍

cla_backend/apps/reports/forms.py
Pheonnexx: This is the same code as the original Case Extract so reluctant to change this for now.

PR #1011, opened 2023-10-27
LGA 2770: Send request body as json

BenMillar-CGI:

- Changes the request body of the `requests.post` method to JSON, this is required as the Orchestrator API is expecting a JSON object.
- Improves the unit testing asserting that the setting value is set correctly
- Removes the default value from `EMAIL_ORCHESTRATOR_URL`, if this is not set and is required i.e. The feature flag is enabled and `TEST_MODE` is not active then it should error.

Intentionally left blank

## Checklist


TawneeOwl (approved):
LGTM 🚀

psweeting1 (approved):
LGTM

PR #1043, opened 2023-12-12
LEP-431 Translate the CFE response

davidread:

Convert the CFE response to the 3 results that is_eligible() normally returns.

### DIFF
--- a/cla_backend/libs/eligibility_calculator/calculator.py
+++ b/cla_backend/libs/eligibility_calculator/calculator.py
@@ -11,6 +11,8 @@ from .cfe_civil.employment import translate_employment
 from .cfe_civil.cfe_response import CfeResponse
 from .cfe_civil.property import translate_property
 
+logger = __import__("logging").getLogger(__name__)
+
 
 class cached_calcs_property(object):
     def __init__(self, func):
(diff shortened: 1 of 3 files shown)


cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -387,6 +388,16 @@ def _translate_case(self):
             request_data.update(translate_employment(self.case_data.you.income, self.case_data.you.deductions))
         return request_data
 
+    def _translate_response(self, cfe_response):
```
starswan, line 391: I had expected this would be a method in CfeResponse. WDYT?

davidread: I guess it's preference. I think OO theory is divided on this subject e.g. https://stackoverflow.com/questions/57009803/object-oriented-design-pattern-for-object-converter

It's just personally I think of a translator to a different type as being a separate thing. CfeResponse is an encapsulation of the response from CFE, and I wouldn't expect to find one of the translators in it.

tbh I'm not a fan of CfeResponse either - rather formal OO for my taste. But can see its methods might be useful.

The tuple response is quite Pythonic, and I guess there's the zen about being explicit - it's really clear this way what it returns, rather than having to dig into the encapsulation to understand that the translation of the response is available.

starswan: I guess if we're returning True, False, 'unknown' then it sits better here where it can be seen by the consumer.

starswan (approved):

PR #1047, opened 2023-12-13
Lep 456 unknown capital check

starswan:

Provide 'incomplete' sections to cfe-civil, so that it can reply 'not_yet_known' if the request is incomplete and the calculation cannot be performed from the given data

Add a local check for 'has_partner' being available (unknown if not available), as CFE cannot be told an 'unknown' value for partner - it defaults to false

## Checklist

### DIFF
--- a/cla_backend/libs/eligibility_calculator/calculator.py
+++ b/cla_backend/libs/eligibility_calculator/calculator.py
@@ -8,12 +8,12 @@ from django.utils import timezone
 from . import constants
 from . import exceptions
 from .cfe_civil.age import translate_age
-from .cfe_civil.dependants import translate_dependants, DEPENDANTS_KEY
-from .cfe_civil.savings import translate_savings, CFE_SAVINGS_KEY
+from .cfe_civil.dependants import translate_dependants, has_dependants_key
+from .cfe_civil.savings import translate_savings, has_savings_key
 from .cfe_civil.employment import translate_employment, has_employment_key
 from .cfe_civil.cfe_response import CfeResponse
-from .cfe_civil.property import translate_property, CFE_PROPERTY_KEY
-from .cfe_civil.income import translate_income, CFE_INCOME_KEY
+from .cfe_civil.property import translate_property, has_property_key
+from .cfe_civil.income import translate_income, has_income_key
 from .cfe_civil.applicant import translate_applicant
 from cla_common.constants import ELIGIBILITY_STATES
 
@@ -438,15 +438,15 @@ class EligibilityChecker(object):
 
     @staticmethod
     def _complete_applicant_data(request_data, facts):
-        return DEPENDANTS_KEY in request_data and hasattr(facts, "has_partner")
+        return has_dependants_key(request_data) and hasattr(facts, "has_partner")
 
     @staticmethod
     def _complete_cfe_income_data(request_data):
-        return has_employment_key(request_data) and CFE_INCOME_KEY in request_data
+        return has_employment_key(request_data) and has_income_key(request_data)
 
     @staticmethod
     def _complete_cfe_capital_data(request_data):
-        return CFE_PROPERTY_KEY in request_data and CFE_SAVINGS_KEY in request_data
+        return has_property_key(request_data) and has_savings_key(request_data)
 
     def _translate_response(self, cfe_response):
         '''Translates CFE-Civil's response to something similar that EligibilityChecker.is_eligible() has always returned'''
--- a/cla_backend/libs/eligibility_calculator/calculator.py
+++ b/cla_backend/libs/eligibility_calculator/calculator.py
@@ -401,10 +401,11 @@ class EligibilityChecker(object):
         if hasattr(case_data, "category"):
             request_data["proceeding_types"] = translate_proceeding_types(case_data.category)
         if hasattr(case_data, "facts"):
-            request_data['applicant'].update(EligibilityChecker._applicant_data(submission_date, case_data.facts))
+            request_data['applicant'].update(
+                EligibilityChecker._translate_applicant_data(submission_date, case_data.facts))
             request_data.update(translate_dependants(submission_date, case_data.facts))
 
-        request_data.update(EligibilityChecker._capital_data(case_data))
+        request_data.update(EligibilityChecker._translate_capital_data(case_data))
         if not EligibilityChecker._complete_cfe_capital_data(request_data):
             request_data['assessment']['section_capital'] = 'incomplete'
 
@@ -440,24 +441,6 @@ class EligibilityChecker(object):
                 return regular_outgoings
         else:
             return regular_income
-
-    @staticmethod
-    def _applicant_data(submission_date, facts):
-        request_data = {}
-        request_data.update(translate_age(submission_date, facts))
-        request_data.update(translate_applicant(facts))
-
-        return request_data
-
-    @staticmethod
-    def _capital_data(case_data):
-        request_data = {}
-        if hasattr(case_data, "you") and hasattr(case_data.you, "savings"):
-            request_data.update(translate_savings(case_data.you.savings))
-
-        if hasattr(case_data, "property_data"):
-            request_data.update(translate_property(case_data.property_data))
-        return request_data
 
     @staticmethod
     def _translate_applicant_data(submission_date, facts):
--- a/cla_backend/libs/eligibility_calculator/calculator.py
+++ b/cla_backend/libs/eligibility_calculator/calculator.py
@@ -396,22 +396,23 @@ class EligibilityChecker(object):
             ]
         }
         if hasattr(case_data, "facts"):
-            request_data['applicant'].update(EligibilityChecker._applicant_data(submission_date, case_data.facts))
+            request_data['applicant'].update(
+                EligibilityChecker._translate_applicant_data(submission_date, case_data.facts))
             request_data.update(translate_dependants(submission_date, case_data.facts))
 
-        request_data.update(EligibilityChecker._capital_data(case_data))
+        request_data.update(EligibilityChecker._translate_capital_data(case_data))
         if not EligibilityChecker._complete_cfe_capital_data(request_data):
             request_data['assessment']['section_capital'] = 'incomplete'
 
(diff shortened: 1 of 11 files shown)


cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -390,24 +399,64 @@ def _translate_case(self):
...
+        return "dependants" in request_data and hasattr(facts, "has_partner")
+
+    @staticmethod
+    def _complete_cfe_income_data(request_data):
+        employment_data = "employment_details" in request_data or "self_employment_details" in request_data
+        income_data = "regular_transactions" in request_data
+        return employment_data and income_data
+
+    @staticmethod
+    def _complete_cfe_capital_data(request_data):
```
davidread, line 455: I don't think this should look at the CFE request, where the schema is clear that these fields are optional. The way we've written the translators happens to put in a blank "properties" section, and others, but that doesn't seem something we should rely on here.

Instead the CaseData object absolutely does define the non-presence of property answers to mean that that the question has not been asked, so I think we should use that.

starswan: My intention is/was that this decision should be 'inside' the translators, so that they produce no output when invalid data is given. Agree the intent could be communicated clearer - I wasn't sure how to indicate that e.g. the 'regular_transactions' key comes from the 'income' translator (and so hence is gross income based) and the 'capitals' key comes from the 'savings' translator and hence is from the 'capital' section

davidread: > e.g. the 'regular_transactions' key comes from the 'income' translator (and so hence is gross income based)

In CFE's schema, `regular_transactions` is for both incomes, like benefits, but it's also for outgoings of types: child_care, rent_or_mortgage, maintenance_out, legal_aid. Whilst CLA doesn't happen to capture those outgoings at the current time, it just makes me confused to see the presence of 'regular_transactions' being used to mean "income is completely filled in on the form". It's flakey. Yes the intent is clearer with CFE_INCOME_KEY. But CLA's native way to tell us if "income is completely filled in in the form" is to check the case_data object for presence of keys. That seems much more straightforward than depending on the output of potentially intertwined translators. But perhaps I'm missing something here.

starswan: I wanted the translators to have this knowledge (which they have to have, as they need to check for existence of properties in general) rather than put it somewhere else. It seemed logical for the translator to not create the key when the fields are not fully populated, and then check for presence afterwards. I've added a tighter abstraction in the last commit to make it clearer - maybe it needs more comments explaining the idea?

davidread: yes this really helps. let's go with this approach 👍

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -353,18 +353,27 @@ def is_eligible(self):
...
+
+            return result, cfe_response
+        else:
+            result = ELIGIBILITY_STATES.UNKNOWN
+            logger.info("Eligibility result (CFE): %s %s" % (result, "couldnt call CFE"))
+            return result, None
 
-    def _translate_case(self, submission_date=None):
+    @staticmethod
+    def _translate_case(case_data, submission_date=None):
```
davidread, line 376: I like that case_data is passed in, and it becomes a staticmethod

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -386,24 +395,63 @@ def _translate_case(self, submission_date=None):
...
+    @staticmethod
+    def _applicant_data(submission_date, facts):
+        request_data = {}
+        request_data.update(translate_age(submission_date, facts))
+        request_data.update(translate_applicant(facts))
+
+        return request_data
+
+    @staticmethod
+    def _capital_data(case_data):
```
davidread, line 432: Let's have "translator" in the name, to indicate that it converts the capital data in case_data to a CFE-request dict.

cla_backend/libs/eligibility_calculator/cfe_civil/savings.py
```diff
@@ -1,25 +1,39 @@
...
+        _savings_value(savings_data, "asset_balance", "Asset")
     ]
+    liquid_capital = none_filter(liquid)
+    non_liquid_capital = none_filter(non_liquid)
 
-    return dict(capitals={
-        "bank_accounts": none_filter(liquid_capital),
-        "non_liquid_capital": none_filter(non_liquid_capital)
-    })
+    if len(liquid_capital) + len(non_liquid_capital) < 3:
```
davidread, line 31: 3 seems very specific, and a source of bugs if we change the number of fields to expect. Either we need a better condition to figure, or simply return `{"back_accounts": [], "non_liquid_capital": []}`

cla_backend/libs/eligibility_calculator/tests/cfe_civil/test_savings.py
```diff
@@ -21,21 +21,13 @@ def test_bank_balance_and_investments(self):
...
-                    "value": 800.11,
+                    "value": 2,
                     "description": "Asset",
                     "subject_matter_of_dispute": False
                 },
-            ]
+
+            ],
         }}
         self.assertEqual(expected, output)
```
davidread, line 33: test_assets seemed useful, to show what happens when some inputs are zero, so not sure why this has been deleted?

cla_backend/libs/eligibility_calculator/tests/cfe_civil/test_savings.py
```diff
@@ -21,21 +21,13 @@ def test_bank_balance_and_investments(self):
...
                 {
-                    "value": 800.11,
+                    "value": 2,
                     "description": "Asset",
                     "subject_matter_of_dispute": False
                 },
-            ]
+
+            ],
         }}
```
davidread, line 32: Can we have a test to see what happens when Savings is empty? i.e. to show this fixes the [empty savings bug]](https://mojdt.slack.com/archives/C04AW468PU4/p1702658658739109):

> On the /about page you say you have savings, and on pressing "Continue" it complains PropertyExpectedException: 'Savings' requires attribute 'bank_balance' and was not given at __init__

i.e.
```
    def test_empty_savings(self):
        savings = Savings()

        output = translate_savings(savings)

        expected = {}
        self.assertEqual(expected, output)
```

cla_backend/libs/eligibility_calculator/cfe_civil/savings.py
```diff
@@ -1,25 +1,40 @@
...
-        _savings_value(savings_data.bank_balance, "Savings"),
-        _savings_value(savings_data.investment_balance, "Investment")
+    liquid = [
+        _savings_value(savings_data, "bank_balance", "Savings"),
+        _savings_value(savings_data, "investment_balance", "Investment")
     ]
-    non_liquid_capital = [
-        _savings_value(savings_data.asset_balance, "Asset")
+    non_liquid = [
+        _savings_value(savings_data, "asset_balance", "Asset")
```
davidread, line 26: Perhaps this could usefully be "Valuable items worth over £500", to reflect the question, rather than the generic "Asset"?

cla_backend/libs/eligibility_calculator/tests/cfe_civil/test_savings.py
```diff
@@ -21,21 +21,21 @@ def test_bank_balance_and_investments(self):
...
-        savings = Savings(bank_balance=0, investment_balance=0, asset_balance=80011)
-        output = translate_savings(savings)
-        expected = {"capitals": {
-            "bank_accounts": [],
             "non_liquid_capital": [
                 {
-                    "value": 800.11,
-                    "description": "Asset",
+                    "value": 2,
+                    "description": "Valuable items worth over 500 pounds",
```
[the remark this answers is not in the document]
davidread: because it's less than the £500?

cla_backend/libs/eligibility_calculator/cfe_civil/savings.py
```diff
@@ -28,7 +28,8 @@ def translate_savings(savings_data):
     liquid_capital = none_filter(liquid)
     non_liquid_capital = none_filter(non_liquid)
 
-    if len(liquid_capital) + len(non_liquid_capital) < 3:
+    # all fields have to be set
```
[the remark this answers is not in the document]
starswan: I've reworked the function now - but if the field were to be renamed, I think we would get test failures as the data would be considered always incomplete

cla_backend/libs/eligibility_calculator/tests/cfe_civil/test_savings.py
starswan: It was deleted due to the mis-understanding about assets_balance being liquid or otherwise, now restored

cla_backend/libs/eligibility_calculator/cfe_civil/savings.py
```diff
@@ -18,23 +19,20 @@ def has_savings_key(request_data):
...
-        _savings_value(savings_data, "asset_balance", "Valuable items worth over 500 pounds")
-    ]
-    liquid_capital = none_filter(liquid)
-    non_liquid_capital = none_filter(non_liquid)
-
-    # all fields have to be set
-    if len(liquid_capital) + len(non_liquid_capital) < len(liquid) + len(non_liquid):
-        return {}
-    else:
+    if has_all_attributes(savings_data, ["bank_balance", "investment_balance", "asset_balance"]):
```
davidread, line 22: The risk of a bug remains - if cla_public renames any of these fields, then this function will silently return nothing. Why so important to return `{}`?

davidread (approved):

cla_backend/libs/eligibility_calculator/cfe_civil/savings.py
starswan: It's the presence of CFE_SAVINGS_KEY - translators return {} to indicate success or failure. Maybe they should return something more clear to indicate this - it's a challenge with the income/deductions anyway as they both return the same CFE key

PR #1050, opened 2023-12-14
LEP-435 refactor

davidread:
Refactor. This way seems marginally more straightforward - what do you think @ripan? I was not a fan of creating a list with Nones in and then a step to remove them.

### DIFF
--- a/cla_backend/libs/eligibility_calculator/cfe_civil/income.py
+++ b/cla_backend/libs/eligibility_calculator/cfe_civil/income.py
@@ -1,27 +1,28 @@
 from cla_backend.libs.eligibility_calculator.cfe_civil.conversions import pence_to_pounds
 
 
+INCOME_CATEGORY_TO_REGULAR_TRANSACTION = {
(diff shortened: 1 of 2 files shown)


ripan (approved):

cla_backend/libs/eligibility_calculator/cfe_civil/income.py
```diff
@@ -15,14 +15,15 @@ def translate_income(income_data):
...
-        if amount_pence:
-            regular_transactions.append(
-                {
-                    "category": INCOME_CATEGORY_TO_REGULAR_TRANSACTION[income_category],
-                    "operation": "credit",
-                    "frequency": "monthly",
-                    "amount": pence_to_pounds(amount_pence)
-                }
-            )
+        if hasattr(income_data, income_category):
```
davidread, line 18: We agree that we don't know why this line is necessary, but it makes the behave tests pass for an unknown reason...

PR #1051, opened 2023-12-14
Submission date is today

davidread:
Previously the date was hard-coded.

The correct date for the calculation would be the day of application, but CLA might be days or weeks in the future. Today seems a reasonable date to choose.

At some point we'll have tests of this function, so I've provided an optional param to fix the date to a known one.

### DIFF
--- a/cla_backend/libs/eligibility_calculator/calculator.py
+++ b/cla_backend/libs/eligibility_calculator/calculator.py
@@ -361,9 +361,10 @@ class EligibilityChecker(object):
 
         return result, cfe_response
 
-    def _translate_case(self):
+    def _translate_case(self, submission_date=None):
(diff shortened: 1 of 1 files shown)


davidread:
@starswan [said](https://github.com/ministryofjustice/cla_backend/pull/1047/files#diff-b8377018daa63cb2c296284d84276674c5801adb824bd151a0b96b68beee2b06R374):
> This should be today, but needs to be in the past so that we can
> compare CFE responses with CLA responses from when it was written (as it uses constants)

Maybe we update any of CLA's constants, if they are out of date. I suspect we'll find the accuracy of them is a real mixture, rather than them all being accurate as of 19/5/2022.

starswan (approved):

PR #1054, opened 2023-12-15
LGA-2799: Announce we’re calling from CLA

BenMillar-CGI:
- Adds a new NullBooleanField to the PersonalDetails model determining whether the call centre operator can announce if they are calling from Civil Legal Advice.

- Adds migrations to add the new field to the database
- Adds 'announce_call' as a field to the personal details unit tests, defaulting to `False`

## Checklist


TawneeOwl (approved):
LGTM

cla_backend/apps/legalaid/models.py
```diff
@@ -158,6 +158,8 @@ class PersonalDetails(CloneModelMixin, TimeStampedModel):
     # only normalised version of post code for now
     search_field = models.TextField(null=True, blank=True, db_index=True)
 
+    announce_call = models.BooleanField(blank=False, default=False)
```
said-moj, line 161: @BenMillar-CGI Should this be a null boolean field? So that old records can have a default value of null

BenMillar-CGI: Hi @said-moj, this implementation defaults all existing records to False, which was the requested behaviour. However, the frontend will display the warning whether the value is null or False. 

I am happy to update this if you think this is the best way to approach it 👍

BenMillar-CGI: Hi @said-moj I have updated this model to use `NullBooleanField` as suggested. Many thanks!! 😄

TawneeOwl (approved):
LGTM

digitalali-moj (approved):
LGTM!

PR #1055, opened 2023-12-15
Lep 455 passported through cfe

starswan:

Enable passported non-partner cases through  CFE rather than legacy_check()

## Checklist

### DIFF
--- a/cla_backend/libs/eligibility_calculator/calculator.py
+++ b/cla_backend/libs/eligibility_calculator/calculator.py
@@ -347,15 +347,39 @@ class EligibilityChecker(object):
         return self.disposable_capital_assets <= limit
 
     def is_eligible(self):
-        cfe_result, cfe_detail = self._do_cfe_civil_check()
+        cfe_result, cfe_response = self._do_cfe_civil_check()
 
-        if self.case_data.facts.on_nass_benefits and self.case_data.category == "immigration":
+        # temp: start doing just non-partner passported checks through CFE
+        if self._is_non_means_tested(self.case_data) or self._is_passported_without_partner(self.case_data):
+            self.calcs = {
+                "pensioner_disregard": int(cfe_response.pensioner_disregard * 100),
+                "disposable_capital_assets": int(cfe_response.disposable_capital_assets * 100),
+                "property_equities": [int(x * 100) for x in cfe_response.property_equities],
+                "property_capital": int(cfe_response.property_capital * 100),
+                "liquid_capital": int(cfe_response.liquid_capital * 100),
+                "gross_income": 0,
+                "partner_allowance": 0,
+                "disposable_income": 0,
+                "dependants_allowance": 0,
+                "employment_allowance": 0,
+                "partner_employment_allowance": 0,
+            }
+
             return cfe_result
         else:
             legacy_result = self._legacy_check()
             logger.info("Eligibility result (legacy): %s %s" % (legacy_result, self.calcs))
 
             return legacy_result
+
+    @staticmethod
+    def _is_passported_without_partner(case_data):
+        return case_data.facts.on_passported_benefits and hasattr(case_data.facts,
+                                                                  "has_partner") and not case_data.facts.has_partner
+
+    @staticmethod
+    def _is_non_means_tested(case_data):
+        return case_data.facts.on_nass_benefits and case_data.category == "immigration"
 
     def _do_cfe_civil_check(self):
         cfe_request_dict = self._translate_case(self.case_data)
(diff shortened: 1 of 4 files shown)


cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -345,26 +345,54 @@ def is_disposable_capital_eligible(self):
         return self.disposable_capital_assets <= limit
 
     def is_eligible(self):
-        self._do_cfe_civil_check()
-
-        legacy_result = self._legacy_check()
-        logger.info("Eligibility result (legacy): %s %s" % (legacy_result, self.calcs))
+        cfe_result, cfe_details = self._do_cfe_civil_check()
```
davidread, line 348: calling the second one `cfe_response` would be clearer to me

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -345,26 +345,54 @@ def is_disposable_capital_eligible(self):
...
+
+        # temp: start doing just non-partner passported checks through CFE
+        if self.case_data.facts.on_passported_benefits and hasattr(self.case_data.facts,
+                                                                   "has_partner") and not self.case_data.facts.has_partner:
```
davidread, line 352: This is how 'black' lays it out, and I think it's an improvement:
```
        if (
            self.case_data.facts.on_passported_benefits
            and hasattr(self.case_data.facts, "has_partner")
            and not self.case_data.facts.has_partner
        ):
```

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -345,26 +345,54 @@ def is_disposable_capital_eligible(self):
...
+                "liquid_capital": int(cfe_details.liquid_capital * 100),
+                "gross_income": 0,
+                "partner_allowance": 0,
+                "disposable_income": 0,
+                "dependants_allowance": 0,
+                "employment_allowance": 0,
+                "partner_employment_allowance": 0,
+            }
+
+            return cfe_result
```
davidread, line 367: Rather than return here, can we both checks, so we can look at the logs during use and (visually at least) compare the outputs of CFE and legacy_check()?

cla_backend/libs/eligibility_calculator/cfe_civil/income.py
```diff
@@ -10,20 +10,28 @@
...
-                    {
-                        "category": INCOME_CATEGORY_TO_REGULAR_TRANSACTION[income_category],
-                        "operation": "credit",
-                        "frequency": "monthly",
-                        "amount": pence_to_pounds(amount_pence)
-                    }
-                )
-    return {"regular_transactions": regular_transactions} if regular_transactions else {}
+            amount_pence = getattr(income_data, income_category)
+            regular_transactions.append(
```
ripan: @starswan We are not going to use regular_transactions with amount=0 anywhere. What if we only add non-zero regular_transactions only?

davidread:
@starswan I changed the base of this PR, so that LEP-456 changes are kept separate. I believe once LEP-456 is merged, then this PR will automatically change its base back to LEP-456.

cla_backend/libs/eligibility_calculator/tests/test_calculator.py
```diff
@@ -931,7 +943,7 @@ def test_pensioner_300k1p_house_100k1p_mort_799999_savings(self):
             "dependants_allowance": 0,
             "employment_allowance": 0,
             "partner_employment_allowance": 0,
-            "liquid_capital": 79999,
+            "liquid_capital": 0,
```
davidread, line 946: interesting! what's going on here?

starswan: This is the test I had to fix - the 'assets' are not liquid capital, despite this test thinking that they are

cla_backend/libs/eligibility_calculator/calculator.py
starswan: I have added some methods so that the layout problems disappear

ripan (approved):

cla_backend/libs/eligibility_calculator/tests/test_calculator.py
davidread: Feels like this bug should be moved to a separate PR, possibly against master

starswan: As the bug is in legacy_check and doesn't affect the calculation, I'm not sure if it is worth fixing?

davidread: Might not be worth it, but I've done a change here: https://github.com/ministryofjustice/cla_backend/pull/1066

PR #1056, opened 2023-12-15
Fix savings translator for empty input

davidread:
Fix for bug in the savings translator - you get exception when sending an empty input: `translate_savings(Savings())`

To reproduce: on the /about page you say you have savings, and on pressing "Continue" it complains:
`PropertyExpectedException: 'Savings' requires attribute 'bank_balance' and was not given at __init__`

As discussed: https://mojdt.slack.com/archives/C04AW468PU4/p1702658658739109

Also this doesn't add empty sections to CFE's request. I think it is clearer. It will throw a spanner into the works of this though: https://github.com/ministryofjustice/cla_backend/pull/1047#discussion_r1428135677

### DIFF
--- a/cla_backend/libs/eligibility_calculator/cfe_civil/savings.py
+++ b/cla_backend/libs/eligibility_calculator/cfe_civil/savings.py
@@ -1,25 +1,32 @@
-from cla_backend.libs.eligibility_calculator.cfe_civil.conversions import pence_to_pounds, none_filter
+from collections import defaultdict
+
+from cla_backend.libs.eligibility_calculator.cfe_civil.conversions import pence_to_pounds
 
 
-def _savings_value(value, description):
-    if value > 0:
+def translate_savings(savings_data):
+
+    def _savings_value(value_pence, description):
         return {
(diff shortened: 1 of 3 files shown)


cla_backend/libs/eligibility_calculator/cfe_civil/savings.py
```diff
@@ -1,25 +1,32 @@
...
+        )
+
+    if hasattr(savings_data, "investment_balance") and savings_data.investment_balance > 0:
+        capitals["bank_accounts"].append(
+            _savings_value(savings_data.investment_balance, "Investment")
+        )
+
+    if hasattr(savings_data, "asset_balance") and savings_data.asset_balance > 0:
+        capitals["non_liquid_capital"].append(
+            _savings_value(savings_data.asset_balance, "Asset")
```
davidread, line 29: Could this be "Valuable items worth over £500" rather than "Asset"?

davidread:
We're going with https://github.com/ministryofjustice/cla_backend/pull/1047

PR #1064, opened 2023-12-20
Rename "liquid_capital" -> "non_property_capital"

davidread:

Rename all occurences of "liquid_capital" -> "non_property_capital"

## Why?

Reasons for rename:
* "liquid capital" doesn't mean what most people think it means. The dictionary says it is "cash"/"bank_balance". But CLA also includes in this calculation result all the assets apart from property, so confusingly includes: "credit" - money owed to the person, "asset" - items worth over £500 and "investment_balance" including stocks, shares, bonds. (This is documented in [one related place](https://github.com/ministryofjustice/cla_backend/blob/master/cla_backend/libs/eligibility_calculator/models.py#L134))
* To avoid confusion with CFE-Civil's output of the same name "liquid_capital", which is just cash/bank balance. This will be important when the integration with CFE is merged.

This changes the cla_backend API responses. It includes the "calculations" field, which is used to expose some of the workings of the calculation, and one of the figures is the renamed field. The "liquid_capital" field is not used outside of cla_backend as far as I can tell - I searched of cla_frontend, cla_public and cla-end-to-end-tests and found only a pair of tests that would need updating to reflect the rename - see https://github.com/ministryofjustice/cla_backend/pull/1064 But I don't have a full understanding of API clients, so would appreciate someone checking this is the case. On this, @said-moj said: "if all our unit/e2e tests pass then I am fine with renaming of the field"

## Checklist


psweeting1 (approved):
LGTM

TawneeOwl (approved):
LGTM

PR #1065, opened 2023-12-20
Bug/fix modsec delete

TawneeOwl:

Add in remove the secrule that blocks deletes from CLA_Backend

Intentionally left blank.

## Checklist

### DIFF
--- a/helm_deploy/cla-backend/templates/ingress_v122.yaml
+++ b/helm_deploy/cla-backend/templates/ingress_v122.yaml
@@ -24,6 +24,7 @@ metadata:
     nginx.ingress.kubernetes.io/modsecurity-snippet: |
       SecRuleEngine On
       SecDefaultAction "phase:2,pass,log,tag:github_team=laa-get-access"
+      SecRuleRemoveById 949110
 spec:
   ingressClassName: "modsec"
   tls:
--- a/helm_deploy/cla-backend/templates/ingress_v122.yaml
+++ b/helm_deploy/cla-backend/templates/ingress_v122.yaml
@@ -23,6 +23,7 @@ metadata:
     nginx.ingress.kubernetes.io/enable-modsecurity: "true"
     nginx.ingress.kubernetes.io/modsecurity-snippet: |
       SecRuleEngine On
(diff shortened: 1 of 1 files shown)


said-moj (dismissed):

helm_deploy/cla-backend/templates/ingress_v122.yaml
```diff
@@ -23,6 +23,8 @@ metadata:
     nginx.ingress.kubernetes.io/enable-modsecurity: "true"
     nginx.ingress.kubernetes.io/modsecurity-snippet: |
       SecRuleEngine On
+      SecDefaultAction "phase:2,pass,log,tag:github_team=laa-get-access"
+      SecRuleRemoveById 949110
```
BenMillar-CGI, line 27: According to [this page](https://developers.cloudflare.com/waf/managed-rules/handle-false-positives/), rule 949110 is "Inbound Anomaly Score Exceeded", are we sure we want to disable this specific rule. As the rule covers a large number of areas.

said-moj: @BenMillar-CGI Should this be 911100 instead?

BenMillar-CGI: @said-moj I think so, yes.
According to: [The ModSecurity Logs](https://logs.cloud-platform.service.justice.gov.uk/_dashboards/app/discover#/doc/b95d8900-dd15-11ed-87c8-170407f57c9c/live_k8s_modsec_ingress-2023.12.20?id=D975DC14-7996-3A42-6373-50206A900077) Rule ID 911100 relates to "Method is not allowed by policy".

TawneeOwl: Hi Ben, this removes the warning of evaluation on the system; the issue is that the users of the backend are all the same(ish) ip address so it'll always flag a delete from a large number of similar IPs

TawneeOwl: Updated to only allow delete method

BenMillar-CGI (approved):
Looks good to me, great stuff! 👍

psweeting1 (approved):
LGTM

PR #1067, opened 2023-12-21
LEP-456 "section_complete" logic is centralized

davidread:
This is an alternative path to https://github.com/ministryofjustice/cla_backend/pull/1062

### DIFF
--- a/cla_backend/libs/eligibility_calculator/tests/test_calculator.py
+++ b/cla_backend/libs/eligibility_calculator/tests/test_calculator.py
@@ -1876,6 +1876,7 @@ class DoCfeCivilCheckTestCase(unittest.TestCase):
                            ),
             'deductions': dict(income_tax=tax, national_insurance=ni)
         })
+        self.income_sections_are_completed(cd)
         case_data = CaseData(**cd)
         return EligibilityChecker(case_data=case_data)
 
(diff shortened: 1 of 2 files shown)


cla_backend/libs/eligibility_calculator/tests/test_calculator.py
```diff
@@ -1975,24 +1959,6 @@ def test_cfe_with_incomplete_property_data_is_unknown(self):
...
-                                                                  child_benefits=500,
-                                                                  deductions={})._do_cfe_civil_check()
-        self.assertEqual("not_yet_known", cfe_result.overall_result)
-
-    def test_incomplete_self_employment_is_unknown(self):
-        _, cfe_result = self.checker_with_income_without_earnings(maintenance_received=100,
-                                                                  child_benefits=500,
-                                                                  self_employed=True)._do_cfe_civil_check()
-        self.assertEqual("not_yet_known", cfe_result.overall_result)
-
```
davidread, line 1995: I've deleted these tests since they aren't realistic situations and unhelpful - if the income section is not complete then maintenance_received property doesn't exist, for example.

davidread: tests added back now!

PR #1075, opened 2024-01-05
Update README install instructions

davidread:
README instructions freshened with a few learnings.

Although VIRTUAL_ENV.md is not our preferred method, and we should not put effort into it, we have made some progress with it and so I thought it reasonable to record learnings, and copy preferred ways to do things from the cla_frontend README.


said-moj (approved):
LGTM, thank you

PR #1080, opened 2024-01-08
LEP-456 Remove older version of section_incomplete logic

davidread:
The older version of this logic (_complete_cfe_capital_data() and _complete_cfe_income_data()) is removed, that is an accidental left-over from when it was replaced in [lep-456-complete-logic-is-centralized](https://github.com/ministryofjustice/cla_backend/pull/1067/files).


starswan:
seems to have broken a couple of tests around over 60 handling?

starswan (approved):

PR #1085, opened 2024-01-08
Fix for deductions being ignored unless all non-zero

davidread:


cla_backend/libs/eligibility_calculator/cfe_civil/deductions.py
```diff
@@ -24,7 +27,9 @@ def translate_deductions(deductions):
                         "amount": pence_to_pounds(amount_pence)
                     }
                 )
-    if len(regular_transactions) != len(_OUTGOING_CATEGORY_TO_REGULAR_TRANSACTION):
-        return {}
+        else:
+            # missing key indicates data is incomplete
+            logger.error("Expected deduction key: %s. Ignoring deductions data!" % outgoing_category)
+            return {}
```
starswan, line 33: think this code should have been deleted rather than moved - its the old incomplete code

davidread: removed!

starswan: There's now a lint failure, but otherwise ok

starswan (approved):

PR #1091, opened 2024-01-09
is_eligible() returns yes/no/unknown

davidread:
## What
is_eligible() returns ELIGIBILITY_STATES - "yes"/"no"/"unknown" instead of True/False/PropertyExpectedException

Screenshot 2024-01-09 at 11 53 21

To fit with this change to the is_eligible() return, the callers have been adjusted:

* is_eligible() is called from apps/legalaid/models.py::EligibilityCheck.get_eligibility_state() when eligibility checks are done from cla_public or cla_frontend
* is_eligible() is also called during some tests with DEBUG=1 from means_test_api

## Why

The reason is that the CFE checker will not naturally raise an exception to signal "unknown" ("not_calculated_yet" in CFE language).

And exceptions that travel several calls up are difficult for new developers to understand, especially when used for expected behaviour on the happy path (rather than an error). You can see why it was used in the past, when the calculation accessed data that was not present, but when the calculation is in CFE this is no longer possible without artificially raising the exception. And there is no need.

## Checklist

### DIFF
--- a/cla_backend/apps/legalaid/models.py
+++ b/cla_backend/apps/legalaid/models.py
@@ -420,19 +420,16 @@ class EligibilityCheck(TimeStampedModel, ValidateModelMixin):
     def get_eligibility_state(self):
         """
         Returns one of the ELIGIBILITY_STATES values depending on if the model
-        is eligible or not. If PropertyExpectedException is raised, it means
-        that we don't have enough data to determine the state so we set the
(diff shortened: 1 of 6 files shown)


said-moj (approved):
👍 LGTM

BenMillar-CGI (approved):
Looks good to me 👍

PR #1108, opened 2024-01-12
LEP-478 Fixup is_eligible() caller to obtain reasons at the same time

starswan:

Convert is_eligible() call to is_eligible_with_reasons() so that the reasons are known by the caller without having to call the checker again or to do some awkward caching of the result

## Checklist

### DIFF
--- a/cla_backend/apps/legalaid/models.py
+++ b/cla_backend/apps/legalaid/models.py
@@ -429,18 +429,14 @@ class EligibilityCheck(TimeStampedModel, ValidateModelMixin):
         ec = EligibilityChecker(case_data)
         eligibility_state, is_gross_income_eligible, is_disposable_income_eligible, is_disposable_capital_eligible = ec.is_eligible_with_reasons()
 
+        reasons = []
         if eligibility_state == ELIGIBILITY_STATES.NO:
-            reasons = []
-
-            def add_reason(boolean, reason):
-                if not boolean:
-                    reasons.append(reason)
-
-            add_reason(is_disposable_capital_eligible, ELIGIBILITY_REASONS.DISPOSABLE_CAPITAL)
-            add_reason(is_gross_income_eligible, ELIGIBILITY_REASONS.GROSS_INCOME)
-            add_reason(is_disposable_income_eligible, ELIGIBILITY_REASONS.DISPOSABLE_INCOME)
-        else:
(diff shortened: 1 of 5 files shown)


cla_backend/libs/eligibility_calculator/cfe_civil/cfe_response.py
```diff
@@ -18,6 +18,23 @@ def employment_allowance(self):
     def gross_upper_threshold(self):
         return self._result_summary["gross_income"]["proceeding_types"][0]["upper_threshold"]
 
+    # The 'is_xxx_eligible checks are actually looking for reasons of ineligibility
+    # hence we check for not 'ineligible' rather than the other way round
```
davidread, line 22: The name of the function "is_gross_eligible" is misleading then, in the case of "not_known_yet" or "not_calculated". How about `is_ineligible_on_gross_income_test`? Or `did_gross_income_test_fail`

On the language, I'm keen that we start calling GI/DI/Capital things "tests". And they are "pass / fail" rather than "eligible" / "ineligible".

davidread: Actually maybe it's a bit surprising these values are inverse - True meaning fail.

How about tri-state? `did_gross_income_test_pass` and return True/False/None - the latter to mean `not_yet_known`/`not_calculated`?

cla_backend/apps/legalaid/models.py
```diff
@@ -428,30 +427,21 @@ def get_eligibility_state(self):
         logger.debug('CaseData %s' % json.dumps(case_data_dict, indent=4, sort_keys=True))
         logger.debug('CaseData is missing: %s' % json.dumps(case_data_dict_missing, indent=4, sort_keys=True))
         ec = EligibilityChecker(case_data)
-        eligibility_state = ec.is_eligible()
+        eligibility_state, is_gross_income_eligible, is_disposable_income_eligible, is_disposable_capital_eligible = ec.is_eligible_with_reasons()
```
davidread, line 430: is it any neater if `is_eligible_with_reasons` returned (eligibilty_state, reasons)?

davidread: seems fine

davidread (approved):

PR #1109, opened 2024-01-12
LGA-2884 Allow access from Global Protect VPN

davidread:

Digital VPN is [disappearing soon](https://mojdt.slack.com/archives/C67Q5TN57/p1698421193318189). 

This PR gives access via Global Protect VPN instead.

IP addresses from GP are listed here:
* https://github.com/ministryofjustice/moj-ip-addresses/blob/main/moj-cidr-addresses.yml#L296
* https://mojdt.slack.com/archives/C06ATC097B5/p1702994991148149

Tested:
Screenshot 2024-01-12 at 18 56 35

This PR doesn't remove access from "MoJ Digital Wifi & Digital VPN". Rather than switch off access from that in this same deployment, I suggest we have a short period of overlap, where both VPNs can be used, to allow all devs to check the new arrangement works for everyone. So removing the "MOJ Digital VPN" IP from this list can be a follow up change.

## Checklist

### DIFF
--- a/helm_deploy/cla-backend/values.yaml
+++ b/helm_deploy/cla-backend/values.yaml
@@ -46,7 +46,7 @@ ingress:
     name: ~
     weight: ~
   whitelist:
-    # MoJ
+    # MoJ Digital Wifi & Digital VPN
     - 81.134.202.29/32
     # MoJ/LAA
     - 52.17.239.55/32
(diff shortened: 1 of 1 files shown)


Pheonnexx (approved):
LGTM 👍

digitalali-moj (approved):
Thanks for sorting this!

PR #1112, opened 2024-01-16
LEP-483 Fix exception for missing income-tax

davidread:
Looks like we forgot to test for the attribute existing before access it.

What keys are present in CaseData depends on what gets stored by the API client, and there are are no controls - so we have to cope with any not being present..

test_missing_income_tax reproduces the exception seen:

    PropertyExpectedException
    Deductions requires attribute income_tax and was not given at __init__
    cla_backend/libs/eligibility_calculator/cfe_civil/employment.py:10

test_missing_national_insurance reproduces a similar one for national insurance.


BenMillar-CGI (approved):
Looks good to me, thank you for resolving this so quickly 👍

TawneeOwl (approved):
LGTM 🚀

psweeting1 (approved):
LGTM

PR #1115, opened 2024-01-19
LEP-485 Refactor CaseData to have defaults, so that translators are no longer vulnerable to PropertyExpectedException gotcha

starswan:

If code accesses a field in CaseData that has not been set, then with this PR, it no longer raises `PropertyExpectedException`. Instead, a default value is returned, which is defined for each field - usually a 0 or False. See: [models.py](https://github.com/ministryofjustice/cla_backend/pull/1115/files#diff-97c27f63e1a4e4fd521bc5afabfdc0123c295ed9d67e23ba2b8fe3581f71f23dR27-R31)

This means that the translators no longer need to preface every access of CaseData fields with a `hasattr()` to check the field exists, or run the risk of the exception being raised.

### Why

Clearly the removal of lots of `hasattr()` simplifies the code and makes it more readable.

But the main benefit is that it removes a coding "gotcha" - if the coder of a translator forgets to add the `hasattr()`, then in the case that the field has not been set by the client, when the translator accesses the field, then 'PropertyExpectedException' is raised, causing the calculation to be aborted.

We had this type of bug, which happened for a case in cla_frontend [and it showed "Server error"](https://mojdt.slack.com/archives/C0637K7CUES/p1705427664930109?thread_ts=1705409829.964429&cid=C0637K7CUES) in one instance. If it were to occur for cla_backend then it would interpret the error as 'unknown', and ask unnecessary questions or refer to the call centre incorrectly.

### How this all works

cla_frontend and cla_public, as clients to cla_backend's API, can set some fields of the CaseData but not others. Which ones are set depends on what data has been collected from the user so far, so depends on how far the user has navigated through the different pages, and depends on the client's particular design of the form, which may show or hide different sections, and require some fields but not others.

cla_backend handles this CaseData and has two jobs:

1. figure out the "completeness" of the data - e.g. the calculation may be complete if all the capital questions have been completed, and likewise for gross income questions and disposable income questions. (This is done in `_is_data_complete_enough_to_call_cfe()`, `_translate_section_gross_income()`, `_translate_section_disposable_income()` and `_translate_section_capital()`) (Ideally the client would tell us this, but it can be divined from which fields are set or not)

2. translate the data to CFE's format - this is the job of the main body of "translators". Any fields which are not present can be defaulted to zero or False, because the calculation has already been told earlier which sections are complete or not.

## Checklist


cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -87,14 +91,11 @@ def _do_cfe_civil_check(self):
...
     def _is_data_complete_enough_to_call_cfe(case_data):
-        if not hasattr(case_data, "facts"):
-            # facts section is at the root of everything, so we require it call CFE
-            return False
         if EligibilityChecker._under_18_passported(case_data):
             # no more info needed
             return True
-        if not hasattr(case_data.facts, "dependants_young") and not (
-            hasattr(case_data.facts, "on_passported_benefits") and case_data.facts.on_passported_benefits
+        if (case_data.facts.on_passported_benefits is None) or (
```
davidread, line 97: This PR changes the outcome for the edge case that `on_passported_benefits=None and dependants_young != None` - this PR changes it to say it can't call CFE. Previously it could call CFE, assuming not on passported benefits, and that seems fine to me - or did you spot a problem with that? If it stays, then we need to explain that situation in the comment and log comment.

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -176,23 +177,22 @@ def is_gross_income_complete(case_data):
...
-            person = case_data.you
-            if not hasattr(person, "deductions"):
-                return False
-            deductions = case_data.you.deductions
-            for key in deductions.PROPERTY_META:
-                if not hasattr(deductions, key):
-                    return False
 
-            if not hasattr(case_data.facts, "has_partner"):
+            if case_data.facts.has_partner and not is_deductions_complete(case_data.partner.deductions):
```
davidread, line 195: We've lost the logic that if `has_partner == None` then we need to return False, for the reason in the comment.

`if case_data.facts.has_partner and not is_deductions_complete(case_data.partner.deductions)` is a welcome improvement, but suggest it is added in addition to the existing logic.

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -176,23 +177,22 @@ def is_gross_income_complete(case_data):
 
     @staticmethod
     def _translate_section_disposable_income(case_data, request_data):
+        def is_deductions_complete(deductions):
+            for key in deductions.PROPERTY_META:
+                if getattr(deductions, key) is None:
+                    return False
+            return True
+
```
davidread, line 185: Again I think the docstring needs to be kept attached to the method

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -216,20 +216,19 @@ def is_property_complete(case_data):
...
+        def is_savings_data_complete(savings):
             for key in savings.PROPERTY_META:
                 if not isinstance(getattr(savings, key), (types.IntType, types.LongType)):
                     return False
-        except exceptions.PropertyExpectedException:
-            return False
-        return True
+            return True
+
+        if case_data.facts.has_partner:
```
davidread, line 226: Same issue about losing the logic of `has_partner` being missing or `== None`

cla_backend/libs/eligibility_calculator/cfe_civil/conversions.py
```diff
@@ -2,8 +2,11 @@
 
 
 def pence_to_pounds(pence):
-    decimal_value = (Decimal(pence) / 100).quantize(Decimal(".01"))
-    return float(decimal_value)
+    if pence is None:
+        return 0
```
davidread, line 6: I'm not a fan of setting a blanket default value of 0 for all monetary figures. Feels more in control if we do it on a per-field basis. When you call this function you could add the default with `or 0`. e.g. in employment.py: `"tax": -pence_to_pounds(deductions.income_tax or 0),`

cla_backend/libs/eligibility_calculator/cfe_civil/employment.py
```diff
@@ -7,30 +7,27 @@
...
-    if hasattr(income, "self_employment_drawings"):
-        gross += pence_to_pounds(income.self_employment_drawings)
+    gross = pence_to_pounds(income.earnings)
+    gross += pence_to_pounds(income.self_employment_drawings)
 
     if (
         gross == 0
-        and (not hasattr(deductions, "income_tax") or deductions.income_tax == 0)
-        and (not hasattr(deductions, "national_insurance") or deductions.national_insurance == 0)
+        and (pence_to_pounds(deductions.income_tax) == 0)
```
davidread, line 24: `and not deductions.income_tax`

davidread:
I rebased, as the diff was confusingly showing bits of other work

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -87,14 +86,11 @@ def _do_cfe_civil_check(self):
...
     def _is_data_complete_enough_to_call_cfe(case_data):
-        if not hasattr(case_data, "facts"):
-            # facts section is at the root of everything, so we require it call CFE
-            return False
         if EligibilityChecker._under_18_passported(case_data):
             # no more info needed
             return True
-        if not hasattr(case_data.facts, "dependants_young") and not (
-            hasattr(case_data.facts, "on_passported_benefits") and case_data.facts.on_passported_benefits
+        if case_data.facts.dependants_young is None and not (
```
davidread, line 92: In the four "is_complete" methods, we're trying to figure out if the front end has collected enough data. The front end is rather inconsistent - sometimes the key is not there, sometimes it sets it to null, etc. I think that is what we should be checking for.

Whereas this code has been changed to evaluate the field, so the subtle runes we get from the frontend are now filtered through the new layer that applies to defaults to the values. It just so happens that this key is defaulted to None, but if that changed then this line would not work.

So I think these four "is complete" methods should not evaluate the keys - they should continue to use hasattr. The defaulting / key evaluation bevhaviour is for the many other translators.

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -131,7 +127,9 @@ def _translate_case(case_data, submission_date=None):
             request_data["assessment"].update(translate_under_18_passported(case_data.facts))
             request_data.update(translate_dependants(submission_date, case_data.facts))
             if hasattr(case_data, "partner") and case_data.facts.should_aggregate_partner:
-                request_data["partner"] = EligibilityChecker._translate_partner_data(case_data.partner, submission_date)
+                request_data["partner"] = EligibilityChecker._translate_partner_data(
+                    case_data.partner, submission_date
+                )
```
davidread, line 132: this is a reformat

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -147,24 +145,22 @@ def _translate_section_gross_income(case_data, request_data):
...
         """
 
-        def is_gross_income_complete(case_data):
-            if not hasattr(case_data, "you"):
-                return False
-            person = case_data.you
-            if not hasattr(person, "income"):
-                return False
-            income = person.income
-
```
davidread, line 157: I appreciate having this code... it explicitly deals with these cases, rather than figure out what evaluating a nested key will default to

cla_backend/libs/eligibility_calculator/calculator.py
davidread: Then `dependants_young` can have default 0, and simplify translate_dependants

cla_backend/libs/eligibility_calculator/cfe_civil/conversions.py
davidread: @starswan

davidread:
I think we need tests for the translators that don't set some of the fields. We have a few, such as:

```
class TestTranslateDeductions(TestCase):
    def test_missing_field(self):
        ...
        deductions = Deductions(maintenance=4545)
        self.assertEqual(expected, translate_deductions(deductions))
```
which tests the defaulting of the values. (in comparison to the happy path test which specifies all the values `Deductions(maintenance=4545, childcare=3737, mortgage=4242, rent=5757, criminal_legalaid_contributions=2424)`)

I think we need similar tests for empty inputs in: test_dependents, test_income, test_under_18_passported. Not sure about property...

cla_backend/libs/eligibility_calculator/calculator.py
starswan: Really like that idea, I've been struggling with the default values

starswan: changed to check __dict__ presence - hasattr() is implemented in terms of getattr() not throwing an exception, and getattr() has been overridden

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -87,14 +86,14 @@ def _do_cfe_civil_check(self):
 
     @staticmethod
     def _is_data_complete_enough_to_call_cfe(case_data):
-        if not hasattr(case_data, "facts"):
+        if not "facts" in case_data.__dict__:
```
davidread, line 89: Looks good! However it's a bit surprising, so how about preceding this with a comment explaining it (and a docstring why not...):
```
# The CFE means calculation doesn't always need all the case_data fields known. Here we check for the presence of certain fields.
# We can't use hasattr(case_data, "<field_name>") to know if a field has been set, because hasattr() under the hood calls getattr(), and we've overridden ModelMixin.__getattr__() to return a default value when the field is absent. So instead, we check for the key in the case_data.__dict__
```

cla_backend/libs/eligibility_calculator/models.py
```diff
@@ -1,34 +1,44 @@
+import types
+
 from . import exceptions
 
 
 class ModelMixin(object):
     PROPERTY_META = None
 
+    @staticmethod
+    def is_a_subclass_of_model_mixin(property_meta_value):
```
davidread, line 10: Feels a bit prominant to have this method first in the class - maybe move it to be last?

cla_backend/libs/eligibility_calculator/models.py
```diff
@@ -1,34 +1,44 @@
...
 
+    # only called if the attribute is not in the class __dict__ (that is done by __get_attribute__)
     def __getattr__(self, kw):
         obj_name = self.__class__.__name__
 
         if kw in self.PROPERTY_META:
-            raise exceptions.PropertyExpectedException(
-                "'{obj_name}' requires attribute '{kw}' and was not given at __init__".format(kw=kw, obj_name=obj_name)
-            )
+            property_meta_value = self.PROPERTY_META[kw]
```
davidread, line 31: I should add a comment explaining the intention:
```
# Attribute has not been set, so we'll return a default value
```

cla_backend/settings/base.py
```diff
@@ -463,8 +463,9 @@ def bank_holidays_cache_adapter_factory():
 
 EMAIL_ORCHESTRATOR_URL = os.environ.get("EMAIL_ORCHESTRATOR_URL")
 
+# This default will be picked up by running tests in an IDE which does not set CFE_HOST
 CFE_URL = (
-    os.environ.get("CFE_HOST", "https://cfe-civil-staging.cloud-platform.service.justice.gov.uk") + "/v6/assessments"
+    os.environ.get("CFE_HOST", "https://main-cfe-civil-uat.cloud-platform.service.justice.gov.uk") + "/v6/assessments"
```
davidread, line 468: Is this related to LEP-485? (I'm not sure CLA team should be using Eligibility Platform team's UAT environment at all)

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -203,32 +202,29 @@ def is_disposable_income_complete(case_data):
...
-        if not hasattr(case_data.facts, "has_partner"):
+        if not "has_partner" in case_data.facts.__dict__:
             # If they have a partner then that may increase assets that they need to delare, so
             # this section is not complete until we clear up if there is a partner
             return False
 
         if len(case_data.property_data) == 0:
             return True
-        for key, value in case_data.property_data[0].iteritems():
+        for key, value in case_data.property_data[0].items():
```
davidread, line 212: Interesting. I thought iteritems() was preferred in python2, because it does it generatively, even though in python3 items() has the generative behaviour too, so is then preferred

cla_backend/libs/eligibility_calculator/cfe_civil/dependants.py
```diff
@@ -12,12 +12,6 @@ def _dependant_aged(todays_date, age, relationship):
...
-    if hasattr(facts, "dependants_young"):
-        children = [_dependant_aged(todays_date, 15, "child_relative") for _ in range(facts.dependants_young)]
-    else:
-        children = []
-    if hasattr(facts, "dependants_old"):
-        adults = [_dependant_aged(todays_date, 17, "adult_relative") for _ in range(facts.dependants_old)]
-    else:
-        adults = []
+    children = [_dependant_aged(todays_date, 15, "child_relative") for _ in range(facts.dependants_young)]
+    adults = [_dependant_aged(todays_date, 17, "adult_relative") for _ in range(facts.dependants_old)]
```
davidread, line 16: nice!

davidread:
Looks great. A few little suggestions, but functionally all reads right. I just want to do a bit of manual testing to double check

cla_backend/libs/eligibility_calculator/calculator.py
starswan: My IDE showed this to me and said it was incompatible with python 3. It seems fine, we only have a handful of items so generator vs list seems pretty unimportant

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -87,14 +86,20 @@ def _do_cfe_civil_check(self):
 
     @staticmethod
     def _is_data_complete_enough_to_call_cfe(case_data):
-        if not hasattr(case_data, "facts"):
+        """
+        The CFE means calculation doesn't always need all the case_data fields known. Here we check for the presence of certain fields.
+        We can't use hasattr(case_data, "<field_name>") to know if a field has been set, because hasattr() under the hood calls getattr(),
```
davidread, line 91: Sorry to niggle, but the docstring is written for callers, which is what L90 is. However this L91-L92 explains implementation, so should instead be a comment

cla_backend/libs/eligibility_calculator/calculator.py
davidread: agreed it's unimportant, but the original is still better, so I'd argue not helpful to change it in this PR

cla_backend/libs/eligibility_calculator/calculator.py
```diff
@@ -203,32 +208,29 @@ def is_disposable_income_complete(case_data):
...
-            savings = case_data.you.savings
-            for key in savings.PROPERTY_META:
-                if not isinstance(getattr(savings, key), (types.IntType, types.LongType)):
-                    return False
-        except exceptions.PropertyExpectedException:
-            return False
+        savings = case_data.you.savings
+        for key in savings.PROPERTY_META:
+            if not isinstance(getattr(savings, key), (types.IntType, types.LongType)):
+                return False
```
davidread, line 233: Testing revealed an issue with this method. When savings has missing keys, we need it to return False. Previously getattr would raise the PropertyExpectedException, however the new behaviour of getattr is to return default values. I think we need:
```
if key not in savings.__dict__:
    return False
if not isinstance(getattr(savings, key), (types.IntType, types.LongType)):
    # usually: getattr(savings, key) == None
    return False
```

davidread:
I manually tested with the [suggested fix](https://github.com/ministryofjustice/cla_backend/pull/1115#pullrequestreview-1865993556), and otherwise is all good

davidread (dismissed):
Looks great to me. Suggest rebasing and open up to CLA team to review

said-moj (approved):

TawneeOwl (approved):
LGTM

PR #1199, opened 2024-06-25
LGA-3156 Allow null for old cases being updated on CHS

digitalali-moj:

Allow null gtm_anon_id for old cases on CHS because it updates case when someone updates personal details and otherwise expects a value for gtm_anon_id, which isn't on there.

Intentionally left blank.

## Checklist


TawneeOwl (approved):
LGTM

said-moj (approved):

PR #1285, opened 2025-05-16
Fix callback capping

BenMillar-MOJ:
### Problem
When the callback capacity was calculated we queried the database with for all cases with callbacks between now and `num_days` later. This means if the calculation was ran at 14:15 on the 1st of January 2025 with `num days = 7` we would look for all callbacks between 14:15 on the 1st of January 2025 and 14:15 on the 7th of January 2025.

However, the callback capacity API returns data on all callbacks between the 1st of January 2025 00:00:00 and the 7th of January 2025 23:59:59 regardless of when in the day the calculation was run.

This resulted in the system allowing callbacks to be booked 7 days in advance, ignoring the callback cap, assuming the callback time was later in the day than when the booking was being made.

### Solution
Ensure we capture callback information for the entire relevant time period by setting the `end_dt` to `23:59:59` of the final day we return information for.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/checker/call_centre_availability.py
+++ b/cla_backend/apps/checker/call_centre_availability.py
@@ -14,7 +14,7 @@ class CheckerOpeningHours(OpeningHours):
         otherwise the capacity will be based from the set of cases in range.
 
         Args:
-            callback_times (List): List callback times
+            callback_times (List): List of callback date times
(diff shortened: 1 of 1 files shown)


willc-work (approved):

TawneeOwl (approved):
LGTM!

PR #1357, opened 2026-02-19
EL-2491: MCC split endpoint

MazOneTwoOne:

If applied , this:
- creates the `/case/{case_reference}/mcc_split` endpoint
- made a base split class, to keep existing logic and functionality 
- moves the `SerializerConfigurationTest` to test folders

- n/a

## Checklist


cla_backend/apps/cla_provider/forms.py
```diff
@@ -154,61 +156,94 @@ def can_provider_deal_with_category(self, category):
...
+    # validate that the case has not been previously split already
+    def validate_case_split_state(self, non_field_errors):
+        if not self.SKIP_IS_PART_OF_SPLIT and self.case.is_part_of_split():
+            non_field_errors.append("This Case has already been split or it has been generated by another Case")
 
-        if non_fields_errors:
-            self._errors[NON_FIELD_ERRORS] = ErrorList(non_fields_errors)
-            return cleaned_data
+    # #### CATEGORY #####
+    def validate_category(self, cleaned_data):
```
aaj-moj: This method is doing the behaviour of the two subclasses of `SplitBaseCaseForm`.

It would be better if you take these two `if`s, move them inside a new `SplitMCCCaseForm.validate_category()`, and move the logic below into a new `SplitCaseForm.validate_category()`.

The call to `validate_category()` in line 244 will be "routed" to the correct subclass. That way, you don't need `SKIP_SAME_CATEGORY` either.

cla_backend/apps/cla_provider/forms.py
```diff
@@ -154,61 +156,94 @@ def can_provider_deal_with_category(self, category):
...
+    def validate_case_has_category(self, non_field_errors):
         if not self.case.eligibility_check or not self.case.eligibility_check.category:
-            non_fields_errors.append("The selected Case doesn't have any category associated.")
+            non_field_errors.append("The selected Case doesn't have any category associated.")
 
-        # validate that the case has not been previously split already
-        if self.case.is_part_of_split():
-            non_fields_errors.append("This Case has already been split or it has been generated by another Case")
+    # validate that the case has not been previously split already
+    def validate_case_split_state(self, non_field_errors):
```
aaj-moj: Same as the other comment. Best to have two methods: `SplitMCCCaseForm.validate_case_split_state()` and `SplitCaseForm.validate_case_split_state()`.

The first one can be a no-op:

```
    class SplitMCCCaseForm(...):
        def validate_case_split_state(self, non_field_errors):
            pass
```

Then, no need for `SKIP_IS_PART_OF_SPLIT` at all.

aaj-moj (changes requested):

aaj-moj (approved):

PR #1389, opened 2026-04-08
LGA-3930: swap `uri` with `path` in the logs to avoid leaking PII

joselopez-moj:

1. Update the `logformat` value from uWSGI to replace `uri` with `path` to avoid sending search queries that may contain sensitive data. 

2. Replace `username` in the logs with `user_id` — aka `instance.pk` 
3. Replace `email` in the logs with `user.pk`

4. I've written a confluence doc where I run through the problem and a proposed solution: https://dsdmoj.atlassian.net/wiki/spaces/laagetaccess/pages/6112772331/How+CLA+backend+logs+PII+data

Intentionally left blank.

## Checklist


cla_backend/apps/cla_auth/views.py
skdejong, line 60: Could we make this log path safe-by-default by removing `USERNAME` from `_get_request_log_extras `and only logging minimal structured metadata here? At the moment the throttling path still appears to log user-derived values, which seems to leave a PII path open.

joselopez-moj: That's greatly spotted! I've removed it from that method so we don't accidentally log the `username`

willc-work (approved):

charlessco (approved):

PR #1400, opened 2026-04-22
EL-2942: Filtering of event logs for MCC's detailed case

aaj-moj:

It adds the ability for the client making the call to get event logs filtered.

Example: `GET /cla_provider/api/v1/case/FA-3465-9114/logs?codes=CASE_VIEWED&codes=MIS&codes=MIS-OOS`

This is necessary for MCC as it will be showing only these specific event logs:
- CASE_VIEWED
- MIS
- MIS-OOS
- MIS-MEANS
- COI
- SPOP
- REOPEN
- REF-INT
- CLSP

For more info and screenshots: https://dsdmoj.atlassian.net/browse/EL-2944

No.

## Checklist

### DIFF
--- a/cla_backend/apps/cla_provider/tests/api/test_log_api.py
+++ b/cla_backend/apps/cla_provider/tests/api/test_log_api.py
@@ -38,7 +38,7 @@ class LogViewSetTestCase(CLAProviderAuthBaseApiTestMixin, LogAPIMixin, APITestCa
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         response_codes = [log["code"] for log in response.data]
         self.assertIn(code_param, response_codes)
-        other_codes_in_response = set(response_codes) - set((code_param,))
+        other_codes_in_response = set(response_codes) - {code_param}
         self.assertEqual(len(other_codes_in_response), 0)
 
     def test_get_filtered_by_multiple_codes_parameter(self):
--- a/cla_backend/apps/cla_provider/tests/api/test_log_api.py
+++ b/cla_backend/apps/cla_provider/tests/api/test_log_api.py
@@ -38,7 +38,7 @@ class LogViewSetTestCase(CLAProviderAuthBaseApiTestMixin, LogAPIMixin, APITestCa
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         response_codes = [log["code"] for log in response.data]
         self.assertIn(code_param, response_codes)
-        other_codes_in_response = set(response_codes) - set({code_param})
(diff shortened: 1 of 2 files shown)


joselopez-moj (dismissed):

cla_backend/apps/cla_provider/tests/api/test_log_api.py
```diff
@@ -22,3 +22,40 @@ def test_get_404_if_not_access_to_case(self):
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        response_codes = [log["code"] for log in response.data]
+        self.assertIn(code_param, response_codes)
+        other_codes_in_response = set(response_codes) - {code_param}
+        self.assertEqual(len(other_codes_in_response), 0)
+
+    def test_get_filtered_by_multiple_codes_parameter(self):
+        """
+        Test that when multiple codes are specified in the 'codes' GET parameter
+        (comma-separated), LogViewSet.get_queryset() filters logs by all specified codes.
```
willc-work: The comment says “comma-separated” but the test does not send a comma-separated string like:

`?codes=CODE1,CODE2`
Instead it sends repeated parameters, can we update the docstring to accurately reflect how the method works.

cla_backend/apps/cla_provider/tests/api/test_log_api.py
```diff
@@ -22,3 +22,40 @@ def test_get_404_if_not_access_to_case(self):
...
+        """
+        codes_param = (self.event_logs[0].code, self.event_logs[1].code)
```
willc-work: the codes are not being specified, can we be specific int he tests i.e test for REOPEN and/or CASE_VIEWED or is there a reason to use the event_logs generated codes for these tests? at the moment I think this just passes 2 identical codes so that `len(codes)` is what we want it to be. I would feel better if these tests used specific allowed codes.

cla_backend/apps/cla_provider/tests/api/test_log_api.py
aaj-moj: I've updated this. 

I started making it comma-separated, but then changed it and forgot about this docstring.

cla_backend/apps/cla_provider/tests/api/test_log_api.py
aaj-moj: I've made the tests more specific. Now the logs are created here, and the codes are also specified here.

willc-work (approved):

joselopez-moj (approved):

PR #1411, opened 2026-05-07
LGA-3852 - Migrate pip-compile from circleci to github actions

said-moj:

Migrate pip-compile from circleci to github actions.

- Previously in circleci. the pip-compile workflow was called but the job did not actually invoke a call to the pip-compile package (it just installed the dependencies)

- Previously requirements/source/requirements-base.in was modified but pip-compile was not run, so you will see changes to requirements/generated/* due to me rerunning pip-compile

- Pin pip-compile hooks to pip 20, otherwise pre-commit will fail due to incompatibility between our legacy dependencies and later version of pip.

- Change pip-compile target language from python3.10 to 2.7 otherwise it will try to install later version of indirect dependencies during the generation of the .txt requirement files.

## Checklist


.github/workflows/workflow.yml
```diff
@@ -9,7 +9,24 @@ permissions:
...
 jobs:
+
+  pip-compile:
+    name: "Pip Compile"
+    runs-on: "ubuntu-latest"
+    container: python:2.7
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd #v6.0.2
+      - name: Install dependencies
+        run: pip install pip-tools pre-commit==1.14.2
```
skdejong: Would it be worth removing `pip-tools` here since the workflow does not directly invoke `pip-compile`? It looks like The `pip-tools` version is already pinned via the pre-commit hook (`rev: 5.5.0` in `.pre-commit-config.yaml`). Therefore I think pre-commit will manage the hook environment itself.

said-moj: @skdejong good spot, pip-tools has been removed from that list of packages now

skdejong (approved):

.github/workflows/workflow.yml
```diff
@@ -9,7 +9,24 @@ permissions:
...
 jobs:
+
+  pip-compile:
+    name: "Pip Compile"
+    runs-on: "ubuntu-latest"
+    container: python:2.7
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd #v6.0.2
+      - name: Install dependencies
+        run: pip install pre-commit==1.14.2
```
charlessco, line 20: Should we use sha pin instead to maintain the security of the repo?

charlessco (approved):

.github/workflows/workflow.yml
said-moj: t seems you can't use a hash when doing pip install via the cli https://github.com/pypa/pip/issues/3257

PR #1423, opened 2026-06-01
LGA-3927 - Add rate limit to ingress

said-moj:

Add rate limit to ingress.

Intentionally left blank.

## Checklist

### DIFF
--- a/helm_deploy/cla-backend/values-uat.yaml
+++ b/helm_deploy/cla-backend/values-uat.yaml
@@ -68,6 +68,6 @@ pdb:
 # Configuration for rate limiting.
 rateLimit:
   rps: 10
-  rpm: ~
+  rpm: null
   burst: 1
-  limitReqStatusCode: 429
+  use429AsStatusCode: true
--- a/helm_deploy/cla-backend/values-uat.yaml
+++ b/helm_deploy/cla-backend/values-uat.yaml
@@ -65,3 +65,9 @@ envVars:
 pdb:
   enabled: false
 
+# Configuration for rate limiting.
+rateLimit:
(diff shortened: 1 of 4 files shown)


joselopez-moj (dismissed):

helm_deploy/cla-backend/values-uat.yaml
```diff
@@ -65,3 +65,9 @@ envVars:
 pdb:
   enabled: false
 
+# Configuration for rate limiting.
+rateLimit:
+  rps: 10
+  rpm: ~
```
willc-work: what does the tilde do in this scenario?

helm_deploy/cla-backend/values-uat.yaml
```diff
@@ -65,3 +65,9 @@ envVars:
...
+  burst: 1
+  limitReqStatusCode: 429
```
willc-work: we dont actually use the value for `limitReqStatusCode` it is more ike a boolean, in that case should this be changed to true/false instead of 429

helm_deploy/cla-backend/values-uat.yaml
said-moj: The tilde means null, I have replaced it with a null value now.

This config only applies to feature branch uat environments

helm_deploy/cla-backend/values-uat.yaml
said-moj: I have changed this to `use429AsStatusCode` with a true value

willc-work (approved):

joselopez-moj (approved):

PR #1429, opened 2026-06-15
Silas - Make matching of emails case insensitive

said-moj:

Make matching of emails case insensitive.

Intentionally left blank.

## Checklist

### DIFF
--- a/cla_backend/apps/cla_auth/authentication.py
+++ b/cla_backend/apps/cla_auth/authentication.py
@@ -171,6 +171,9 @@ class EntraAccessTokenAuthentication(authentication.BaseAuthentication):
 
         try:
             user = authenticate(entra_id_email=email)
+        except User.MultipleObjectsReturned:
+            logger.error("Multiple users found with the email address  %s", email, exc_info=True)
+            raise exceptions.AuthenticationFailed("Multiple users found with the email address")
         except Exception:
             user = None
 
--- a/cla_backend/apps/cla_auth/authentication.py
+++ b/cla_backend/apps/cla_auth/authentication.py
@@ -172,7 +172,7 @@ class EntraAccessTokenAuthentication(authentication.BaseAuthentication):
         try:
             user = authenticate(entra_id_email=email)
         except User.MultipleObjectsReturned:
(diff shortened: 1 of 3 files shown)


joselopez-moj (dismissed):

cla_backend/apps/cla_auth/backend.py
```diff
@@ -5,7 +5,7 @@
 class EntraAccessTokenAuthenticationBackend(ModelBackend):
     def authenticate(self, entra_id_email=None):
         try:
-            user = User.objects.get(email=entra_id_email, is_active=True)
+            user = User.objects.get(email__iexact=entra_id_email, is_active=True)
```
skdejong: This looks good and the test covers the intended behaviour. One possible edge case: if duplicate active users exist with emails differing only by case, `email__iexact` could raise `MultipleObjectsReturned` and authentication would fail silently due to the broad `except Exception`. Are emails guaranteed unique case-insensitively elsewhere?

said-moj: @skdejong I think the edge case of `MultipleObjectsReturned` silently failing is a valid edge case that we need to consider. However if two users have the same email address but in different casing then we consider those two having the same email address.

skdejong (dismissed):

joselopez-moj (dismissed):

cla_backend/apps/cla_auth/authentication.py
```diff
@@ -171,6 +171,9 @@ def get_or_create_user(self, payload):
 
         try:
             user = authenticate(entra_id_email=email)
+        except User.MultipleObjectsReturned:
+            logger.error("Multiple users found with the email address  %s", email, exc_info=True)
```
skdejong: Is it worth changing this line to 

```
logger.exception(
    "Multiple users found with the email address %s",
    email
)
```

in order to satisfy Sonar and make the intent clearer?

said-moj: @skdejong this has been added now.

btw logger.exception calls logger.error with `exc_info=True`

skdejong (approved):

joselopez-moj (approved):

